PR #8, opened 2020-09-08
StackAllocator

Schtekt:
The stackallocator allocates memory and frees it in a LIFO (Last in First out) manner.


jakkeez:
Tried to merge with master, not allowed in current state

PR #10, opened 2020-09-08
Fonts folder and classes

fredrix96:
A new resource folder has been added called Fonts. This folder is supposed to host all the different fonts we want to use in the game. 
A new class called Text has been added to Renderer, where some necessary structs have been declared. 
A TextTask class and two new shaders called TextVertex and TextPixel have also been added.


Schtekt (commented):
Missing an inclusion of "stdafx.h" in Text.cpp. Looks good otherwise.

jakkeez (changes requested):
Missing an inclusion of "stdafx.h" in Text.cpp. Looks good otherwise. // Tim

jakkeez (approved):

Schtekt (approved):

PR #11, opened 2020-09-08
Random generator

Schtekt:
A random generator class is added. Some tests to show performance exist in Enginetests project.

This is tied to two tasks in User story ID #31.


LeoWikstrom (approved):
Ser bra ut

fjz345 (approved):
float EngineRand::Randf() -----

float nr = static_cast<float>(Rand()) / 10000;

Could be useful adding a comment or variable for the 10000.

PR #12, opened 2020-09-09
Improved texture loading, Loading FBX works

Whixx:
Some textures still doesn't load correctly. Probably because we don't have support for embedded textures.


fjz345 (dismissed):
William added documentation about dragon licence

AntonAsbrink (commented):
There is multiple models and textures added with no example of them working.
I don't know if they are added only as assets for later, but in that case save it for then.

Additionally, try to keep all resources in an appropriate folder. There are floor textures in a folder called dragon.

AntonAsbrink (approved):

fjz345 (approved):

PR #16, opened 2020-09-10
Network Connection

AntonAsbrink:
This feature only adds ability to connect to another ip.
Can be tested by opening two windows and doing following:
Server: Write anything but "1" and it will start listening.
Client: Write "1" + enter and then "localhost" in order to connect to the other application.
Connection can be confirmed by looking at the logs on VS output.


StortSpelProjekt/EngineTests/src/test.cpp
```diff
@@ -362,15 +362,15 @@ TEST(MEMORYMANAGER, HEAPLARGE)
...
-	if(smallMem)
-		MemoryManager::FreeHeapBlock(smallMem);
+{
+	void* smallMem = MemoryManager::AllocHeapBlock();
+	void* smallMemSec = MemoryManager::AllocHeapBlock();
+
+	EXPECT_NE(smallMem, nullptr);
+	EXPECT_NE(smallMemSec, nullptr);
+	
+	if(smallMem)
```
jakkeez, line 372: if-statements blocks are mandatory according to code standard. Required change here.

jakkeez (changes requested):

StortSpelProjekt/EngineTests/src/test.cpp
```diff
@@ -362,15 +362,15 @@ TEST(MEMORYMANAGER, HEAPLARGE)
...
+		MemoryManager::FreeHeapBlock(smallMem);
```
jakkeez, line 373: if-statement blocks are mandatory according to code standard. Change this to a block.

StortSpelProjekt/Server/src/main.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+	if (ip == "1") {
+		std::cin >> ip;
+		std::cout << "Connecting to " + ip << std::endl;
+		network.ConnectToIP(ip, 55555);
+	}
+	else {
+		std::cout << "Listening for connections" << std::endl;
+		network.ListenConnection(55555);
+	}
```
jakkeez: if-statement structure has to be the following:  
if (true)
{
	//do something
}

Changes these statements to reflect this.

jakkeez (changes requested):

jakkeez (approved):

Schtekt (approved):

PR #17, opened 2020-09-10
Player cube created (in Sandbox) + update of camera variable names

LeoWikstrom:
Huge feature. There is now a cube representing the player. The cube does nothing. The misnamed  "lookAt" variables and functions in the camera classes are now called "direction".


Whixx (dismissed):
The player cube seems to have some wrong UVs, but this shouldn't matter since it's just for testing anyway.
Else all seems good.

Schtekt (dismissed):
Removal of the player cube causes the application to crash. I feel that this is an underlying problem in the engine and therefore this commit is approved.

Whixx (approved):

Schtekt (approved):

PR #20, opened 2020-09-10
Errorprinting

Schtekt:
Added error prints to file. User may print text, print text with severity AND reset the error text file.


LeoWikstrom (approved):
Everything seems to work

jakkeez (approved):

PR #23, opened 2020-09-11
Send/Receive packets

AntonAsbrink:
A bare bone implementation allowing to send string packets. Mostly a proof of concept. !OBS only tested for local machine


jakkeez:
Pull request to master not allowed

PR #26, opened 2020-09-11
OBB Collision

OtakuFreak13:
BoundingBoxComponent class now uses DirectX::BoundingOrientedBox to make the bounding box. 
The newly made Physics class then has a collision check that takes in two DirectX::BoundingOrientedBox and returns true if they collide.
Test for collision exists at the bottom of sandbox main: 
The stone object is rotating towards a box object and Log will print out when they collide. Player object can also collide with the box and will print out as well. Note collision detection is only enabled for these specific cases. E.g player and stone won't check collision if they collide.


Whixx (changes requested):
Further testing required, colisions not currently working properly

Whixx (dismissed):
Possibly look at making deconstructors virtual, also don't forget to remove testing.

StortSpelProjekt/Engine/src/ECS/Components/BoundingBoxComponent.h
```diff
@@ -11,33 +22,39 @@ namespace component
...
 	private:
+		// used for collision checks
+		DirectX::BoundingOrientedBox m_OrientedBoundingBox;
+		// inital state of the OBB, used for math in update()
+		DirectX::BoundingOrientedBox m_OriginalBoundingBox;
+		Transform* m_pTransform = nullptr;
+		// If picking and or collision should be enabled
+		unsigned int m_FlagOBB = 0;
+		Mesh* m_pMesh = nullptr;
 		std::string m_pPathOfModel = "";
```
LeoWikstrom: Not a pointer, no 'p' needed: m_PathOfModel

StortSpelProjekt/Engine/src/Physics/Physics.h
```diff
@@ -0,0 +1,19 @@
...
+#include<DirectXCollision.h>
+
+class Physics
+{
+public:
+	Physics();
+	~Physics();
+
+	// returns true if there is an intersection between the OBBs
+	bool checkOBBCollision(DirectX::BoundingOrientedBox obb1, DirectX::BoundingOrientedBox obb2);
```
LeoWikstrom: Public method should start with capital letter: Check

LeoWikstrom (changes requested):
Fix the code standard comments and remove the test and we're good to go

LeoWikstrom (approved):

Whixx (approved):

PR #29, opened 2020-09-14
Bloom

jocke1995:
I implemented bloom and added an overall structure for pp-effects.
Since we don't have any shiny objects in the scene at the moment, I added a temporary solution that increases the emissive color of each object in the scene.


StortSpelProjekt/Engine/src/Renderer/HLSL/ForwardPixel.hlsl
```diff
@@ -74,7 +80,21 @@ float4 PS_main(VS_OUT input) : SV_TARGET0
 			normal.rgb);
 	}
 
-	finalColor += emissiveMap.rgb;
+	// This is a test to show that bloom works, remove after review
+	finalColor += float4(1.0f, 1.0f, 1.0f, 1.0f);// emissiveMap.rgb;
```
jocke1995: The test is here, remove this after the glow effect is confirmed.

StortSpelProjekt/Engine/src/Renderer/DX12Tasks/ForwardRenderTask.cpp
```diff
@@ -54,21 +60,35 @@ void FowardRenderTask::Execute()
 	DescriptorHeap* renderTargetHeap = m_DescriptorHeaps[DESCRIPTOR_HEAP_TYPE::RTV];
 	DescriptorHeap* depthBufferHeap  = m_DescriptorHeaps[DESCRIPTOR_HEAP_TYPE::DSV];
 
-	D3D12_CPU_DESCRIPTOR_HANDLE cdh = renderTargetHeap->GetCPUHeapAt(m_BackBufferIndex);
+	// RenderTargets
+	const unsigned int SwapChainIndex = swapChainRenderTarget->GetDescriptorHeapIndex();
```
Whixx: Capital letter on this variable, is this intended? Else change this for consistency.

Whixx (changes requested):
A variable has a capital letter, i've made a comment for this, look at it and change if it's not intended.
Also remove test.
The Renderer.cpp file has probably had the line endings changed. This is a problem when reviewing since i can't see which places have actually been changed. Look at finding a way to avoid changing the line endings in the future.
Otherwise everything looks fine.

fredrix96 (changes requested):
Change m_resources to m_Resources in file Bloom.h at row 28.
Variables such as timeToBlur in file BlurComputeTask.cpp at row 67 should be created and controlled externally.
Change m_threadsPerGroup to m_ThreadsPerGroup in file BlurComputeTask.h at row 28.
The test case may be a little bit too extreme and may disturb/annoy others if it is merged into develop.
The program runs at it should with no memory leaks or crashes. Good job!

Whixx (dismissed):

Whixx (approved):

fredrix96 (approved):
Looks good!

PR #31, opened 2020-09-14
Audio Looping

AntonAsbrink:
This adds feature to change the amount of times audio loops.
Sandbox contains an audio example for testing this.
This push also contains some changes to Option that was used.


Schtekt (approved):
Looks good, sounds great!

kattalainen (approved):

PR #32, opened 2020-09-14
Implemented WASD movement

LeoWikstrom:


Whixx (changes requested):
Seems to work fine, code looks good.
There is a bug which makes the box stutter slightly when we move it, look into this.

Whixx (approved):
Smooooooooooooooooooooth

OtakuFreak13 (approved):
Ser bra ut

PR #35, opened 2020-09-14
Physics now checks from collision with events

OtakuFreak13:
Physics now checks collision every 0.2 seconds and publishes a Collision event when collision happens. 
Physics also prints out which entities have collided which it gets from the Collision event.


AntonAsbrink (dismissed):

Schtekt (changes requested):
name change for m_Physics in Engine required. needs a p prefix after "m_".
Is it necessary for physics to be allocated on the heap?

Schtekt (approved):

Schtekt:
Edited the changes myself. Looking good now!

AntonAsbrink (approved):

PR #45, opened 2020-09-16
Removal of Texts

fredrix96:
The texts can now be removed and you are now able to name texts to be able to find them more easily. The GPU memory optimization has to wait until we get a better system to handle it


OtakuFreak13 (approved):
Tested and it removes text.  Code also seems to be okey

Schtekt (approved):
Textcomponent pretty straightforward. "Removal of text" is essentially removal of entity, but it is A okay.

PR #54, opened 2020-09-17
SplitRenderComponents

jocke1995:


StortSpelProjekt/Engine/src/Headers/Core.h
```diff
@@ -105,9 +105,11 @@ enum class CAMERA_TYPE
...
 enum FLAG_DRAW
 {
-	ForwardRendering = BIT(1),
-	Blend = BIT(2),
-	Shadow = BIT(3),
+	No_Depth = BIT(1),
+	ForwardRendering = BIT(2),
+	Blend = BIT(3),
+	Shadow = BIT(4),
+	NUM_FLAG_DRAWS = 4,
```
OtakuFreak13: Code Standard says Enums shoud be upper case and with _ and no camelCase

OtakuFreak13 (changes requested):

StortSpelProjekt/Engine/src/Headers/Core.h
OtakuFreak13: So last line 112 is fine but the others need changing

Whixx (dismissed):
Good.

OtakuFreak13 (approved):

Whixx (approved):

PR #60, opened 2020-09-22
Player mouse input

jakkeez:
Added mouse support for left and right click but utilizing events and added a function in the playerinputcomponent that checks which button is pressed and prints a message depending on which button is clicked.


AntonAsbrink (dismissed):

kattalainen (dismissed):

LeoWikstrom (approved):

Schtekt (approved):
Looks good. Nice to see someone else work on premake ;)

PR #62, opened 2020-09-22
Full HD Window plus Fullscreen

fredrix96:
Test if the screen is displayed correctly and if fullscreen works by changing the variables in main.cc


kattalainen (dismissed):

jakkeez (changes requested):
Fix main.cpp in Game. Engine.Init has missing arguments like fullscreen, width and height

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -1096,7 +1096,7 @@ void Renderer::initRenderTasks()
 		COMMAND_INTERFACE_TYPE::DIRECT_TYPE,
 		m_pBloomResources->GetPingPongResource(0),
 		m_pBloomResources->GetPingPongResource(1),
-		800, 600); // TODO: Send screen width/height from window after merge to develop
+		window->GetScreenWidth(), window->GetScreenHeight());
```
jocke1995: Nice!

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -628,7 +628,7 @@ void Renderer::updateMousePicker()
 	}
 }
 
-void Renderer::initRenderTasks()
+void Renderer::initRenderTasks(const Window* window)
```
jocke1995: Isn't window saved as a member variable in Renderer?

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -6,9 +6,14 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int nCmdShow)
 {
     _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
 
+	// Window values (will be influenced by the option class)
+	bool fullscreen = false;
+	int screenWidth = 1920;
+	int screenHeight = 1080;
```
jocke1995: When you change this, you modify both the window size + the resolution of the renderTargets. Is this what you meant to do? This means that we will not be able to test the game in "high-res" without going fullscreen.

jocke1995 (changes requested):
I left a few comments regarding some issues that I might not understand, or that you could possibly have missed.

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
fredrix96: No?

StortSpelProjekt/Game/src/main.cpp
fredrix96: Yes, that is what I meant to do. However, you do have a valid point and I will add a resolutionWidth and a resolutionHeight as well

PR #63, opened 2020-09-22
Acceleration component

Schtekt:
Acceleration implemented in Game project.

A test exists in the default scene to show off gravity. This could or maybe even should be removed before merge.

Note that the "player" bounces a bit in the ground when landing. This is because of how the physical updates differ from the rendering updates. The Renderer manages to draw the object before the physical position has been applied to the rendering position.


StortSpelProjekt/Game/src/Components/AccelrationComponent.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void component::AccelrationComponent::SetAccelrationDirection(DirectX::XMFLOAT3 dir)
+{
+	DirectX::XMStoreFloat3(&m_AccDir, DirectX::XMVector3Normalize(DirectX::XMLoadFloat3(&dir)));
+}
+
+void component::AccelrationComponent::SetAccelrationSpeed(float speed)
+{
+	m_AccSpeed = speed;
+}
```
jocke1995: Misspelling of AccelerationComponent?

StortSpelProjekt/Game/src/Components/AccelrationComponent.h
```diff
@@ -0,0 +1,29 @@
+#ifndef ACCELRATIONCOMPONENT_H
+#define ACCELRATIONCOMPONENT_H
```
jocke1995: Spelling is wrong

jocke1995 (dismissed):
Well done! I spotted a misspeling of "Acceleration", Change that, remove your test in main and then merge.

jocke1995 (approved):
Well done

jakkeez (approved):

PR #64, opened 2020-09-23
EnemyHandler

OtakuFreak13:
Made an enemy handler in Game that makes it easy to add enemies to a scene.
The Handler saves the information of an enemy type to a struct the first time that type is added to a scene. This struct is then used to make the enemy the following times with only enemy name and position needed as input.


StortSpelProjekt/Game/src/Gamefiles/EnemyHandler.h
```diff
@@ -0,0 +1,59 @@
...
+	// to skipp scale and write to rot the equvalent would be FLT_MAX
+	// OBS "pos" goes BEFORE "entityName" in this overloaded function
+	Entity* AddEnemy(float3 pos, std::string entityName, unsigned int flag = UINT_MAX, float scale = FLT_MAX, float3 rot = { FLT_MAX , FLT_MAX , FLT_MAX });
+
+
+
+
+private:
+	Scene* m_pScene = nullptr;
+	std::map<std::string, EnemyComps*> m_enemyComps;
```
Whixx: Stor bokstav i m_enemyComps

Whixx (changes requested):
Look at using more clear names for variables and classes which describe what the actually do. Mainly i want you to think about the name "EnemyHandler", which functions of the enemies do this class actually handle? F;r example if the task of the class is just to produce new enemies i'd say "EnemyFactory" or something similar is more fitting.
Also have another look through the code for codestandard reasons. I've made a comment on a member variable that is not correct.

StortSpelProjekt/Game/src/Gamefiles/EnemyHandler.cpp
```diff
@@ -0,0 +1,183 @@
...
+				Physics::GetInstance().AddCollisionEntity(ent);
+			}
+			return ent;
+		}
+		else
+		{
+			Log::PrintSeverity(Log::Severity::WARNING, "Inssuficient input in parameters to add new type of enemy!\n");
+			return nullptr;
+		}
+	}
```
jocke1995: The naming of the three overloaded functions are a bit confusing, change according to what we discussed in Discord.

StortSpelProjekt/Game/src/Gamefiles/EnemyHandler.h
jocke1995: m_EnemyComps

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Engine.h"
 #include "Components/PlayerInputComponent.h"
+#include "EnemyHandler.h"
```
jocke1995: Should this be included here? This might be the case, but maybe it should be included somewhere in "game" ?

jocke1995 (changes requested):
Change according to my comments

StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.h
```diff
@@ -0,0 +1,57 @@
+#ifndef ENEMY_HANDLER_H
+#define ENEMY_HANDLER_H
```
jocke1995: Should be ENEMYFACTORY_H

jocke1995 (changes requested):
Just a small naming bug from when you changed the name of the class.

Whixx (approved):

jocke1995 (approved):

PR #66, opened 2020-09-23
Resolution, Window size and Config

fredrix96:
The resolution and the window size are now treated as two separate things. You can change these two in the config file. Also, you can now change if you want fullscreen or not. However, an extern fullscreen mode is difficult to get right in DX12. This means that we don't have that, but rather a windowed borderless fullscreen.


jocke1995 (approved):
gj

Whixx (approved):
Well

PR #67, opened 2020-09-23
The program will now run with a maximum of four threads

fredrix96:
The thread limit is set in Engine.h


jocke1995 (dismissed):
It is fine, but I thinks that it is enough to hardcode "maxNumThreads" since the only usage is for us developers before launch.

StortSpelProjekt/Engine/src/Engine.cpp
```diff
@@ -31,9 +31,16 @@ void Engine::Init(HINSTANCE hInstance, int nCmdShow)
...
+	int numThreads = std::thread::hardware_concurrency();
+	if (numThreads == 0) // function not supported
+	{
+		numThreads = 1;
+	}
+	else if (numThreads > m_ThreadLimit) // Limiting the number of threads to the threadLimit
+	{
+		numThreads = m_ThreadLimit;
+	}
+	m_pThreadPool = new ThreadPool(numThreads); // Set num m_Threads to number of cores of the cpu
```
OtakuFreak13: Change the comment on line 43 so it reflects the new code.

OtakuFreak13 (changes requested):
Seems fine just change the comment I mentioned then we're good.

OtakuFreak13 (approved):

jocke1995 (approved):

PR #70, opened 2020-09-24
Material added

fjz345:
Also changed a lot of std::string to std::wstring.
Also changed some arguments to be const& so that you don't have to create a variable when for example inputting a path.


StortSpelProjekt/Engine/src/Misc/AssetLoader.h
```diff
@@ -63,6 +64,7 @@ class AssetLoader
     // Every model & texture also has a bool which indicates if its data is on the GPU or not
     std::map<std::wstring, std::pair<bool, Model*>> m_LoadedModels;
     std::vector<Mesh*> m_LoadedMeshes;
+    std::map<std::wstring, std::pair<bool, Material*>> m_LoadedMaterials;
```
jocke1995, line 67: nice!

jocke1995 (dismissed):

AntonAsbrink (approved):

jocke1995 (approved):

PR #73, opened 2020-09-25
Implemented in-game console using ImGui

LeoWikstrom:
Try the command BOUNDINGBOX


OtakuFreak13 (approved):
Tested both Game and Sanbox and does what it should. Saw no fault with the code.

kattalainen (approved):

PR #75, opened 2020-09-28
SetScene-AddComponentRework

kattalainen:
Changed the way SetScene adds components. Instead of loops, now each component will call on InitScene() to do what they need. Renderer is changed into a singleton for consistency with other sub-engines and ease of access for the components that need it.


OtakuFreak13 (changes requested):
Looks good. Just change the funtion name in Game from the old SetSceneToDraw to the new SetScene

OtakuFreak13 (dismissed):

LeoWikstrom (approved):

OtakuFreak13 (approved):

PR #76, opened 2020-09-28
Simple AI

LeoWikstrom:


kattalainen (approved):

OtakuFreak13 (approved):
Looks good. Really like the jumping attack they do. Merging with develop shouldn't have any conflics since the only difference is in that I added enemies to the demo scene.

PR #81, opened 2020-10-01
NavQuad

AntonAsbrink:
This feature has created a system for navigation through quad and their connections. There is no use for this yet, but a test is in "AntonTestScene"


Schtekt (approved):

jakkeez:
No work for current sprint, will reopen for next sprint

PR #83, opened 2020-10-01
Server Fix

AntonAsbrink:
There were problems with threads in server following an update. It fixed now


jakkeez (approved):

StortSpelProjekt/Server/src/ConsoleCommand.cpp
```diff
@@ -2,7 +2,7 @@
 #include <iostream>
 
 
-Console::Console()
+Console::Console() : MultiThreadedTask(FLAG_THREAD::NETWORK)
```
jocke1995, line 5: Nice

jocke1995 (approved):
yes

PR #87, opened 2020-10-02
It now possible to switch between windowed and fullscreen

fredrix96:
It is done if you can play in fullscreen and if you can alt+tab or alt+enter without any crashes


StortSpelProjekt/Engine/src/Misc/Window.cpp
```diff
@@ -1,28 +1,50 @@
...
+
 	if (ImGui_ImplWin32_WndProcHandler(hWnd, msg, wParam, lParam))
+	{
 		return true;
+	}
 
 	switch (msg)
 	{
+	case WM_SYSKEYDOWN: // alt+enter
+		if (wParam == VK_RETURN && (lParam & 0x60000000) == 0x20000000)
```
jocke1995, line 29: Why do this hack with bitmask-and? Why can't you just use VK_RETURN && VK_MENU?

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
```diff
@@ -28,12 +28,34 @@ class SwapChain
...
+		ID3D12CommandQueue* commandQueue,
+		DescriptorHeap* descriptorHeap_RTV,
+		DescriptorHeap* descriptorHeap_CBV_UAV_SRV);
 
 private:
 	std::array<Resource*, NUM_SWAP_BUFFERS> m_Resources;
 	std::array<RenderTargetView*, NUM_SWAP_BUFFERS> m_RTVs;
 	std::array<ShaderResourceView*, NUM_SWAP_BUFFERS> m_SRVs;
 	IDXGISwapChain4* m_pSwapChain4 = nullptr;
+	DXGI_SWAP_CHAIN_DESC1 scDesc;
```
jocke1995, line 43: Allways init structs with "= {}"

jocke1995: init with "= {}"

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
```diff
@@ -28,12 +28,34 @@ class SwapChain
...
+	bool m_Fullscreen = false;
+
+	// colour format
+	DXGI_FORMAT m_DesiredColourFormat = DXGI_FORMAT_R16G16B16A16_FLOAT;
+	unsigned int m_NumberOfSupportedModes = 0;	// the number of supported screen modes for the desired colour format
+	DXGI_MODE_DESC* m_pSupportedModes = {};		// list of all supported screen modes for the desired colour format
```
jocke1995: = nullptr

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
```diff
@@ -28,12 +28,34 @@ class SwapChain
...
+	DXGI_MODE_DESC m_CurrentModeDescription = {};	// description of the currently active screen mode
+
+	const void initFullscreen(unsigned int* width, unsigned int* height);
+	const void resize(unsigned int width, unsigned int height);
+	const void createSwapBuffers(ID3D12Device5* device,
+		unsigned int width, unsigned int height,
+		DescriptorHeap* descriptorHeap_RTV,
+		DescriptorHeap* descriptorHeap_CBV_UAV_SRV);
+	const void clearSwapBuffers() const;
```
jocke1995: makes no sense to have this function marked as const because it is supposed to "clearSwapBuffers"? Which means that it should change something inside the class? Also why is the return value marked as const? If the return value is void it wont return anything anyways?

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -33,30 +33,30 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int nCmdShow)
...
+			    
+			    renderer->Update(updateRate);
+			    Physics::GetInstance().Update(updateRate);
+			}
+
+			/* ------ Sort ------ */
+			renderer->SortObjects();
+
+			/* ------ Draw ------ */
+			renderer->Execute();
```
jocke1995: Tab by misstake?

StortSpelProjekt/Vendor/config.txt
```diff
@@ -3,10 +3,12 @@ f_backgroundVolume 1.0
...
 i_resolutionHeight 1080
-i_windowWidth 1280
-i_windowHeight 720
-i_fullscreen 1
-i_amountOfBlur 2
+i_windowWidth 1920
+i_windowHeight 1080
+b_fullscreen 0
+b_windowedFullscreen 0
+b_stretchedWindow 0
```
jocke1995: Agreed to make this into a single variable, so that only one of them can be set at the same time

jocke1995 (changes requested):
It runs for me, alt+tab works, gj! Found a few programming issues that I marked in the review.

StortSpelProjekt/Engine/src/Misc/Window.cpp
fredrix96: Because then you need to first look for the first key and then wait until the second key is pressed. WM_SYSKEYDOWN look if alt (or F10) is pressed combined with another key. But, we do not want to do anything unless alt+enter is pressed. And that is what the if-statement is looking for. WM_SYSKEYDOWN is reacting on special events with the help of alt, for example alt+enter which triggers a change from fullscreen to windowed or the other way around. But there are some other important alt combinations as well, but they are not important right now.

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
fredrix96: sure

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
fredrix96: sure

StortSpelProjekt/Engine/src/Renderer/SwapChain.h
fredrix96: Yes, that is a valid point. I do not know why I did that

StortSpelProjekt/Game/src/main.cpp
fredrix96: Github problems, nothing to worry about

StortSpelProjekt/Vendor/config.txt
fredrix96: Yes we did

jocke1995 (dismissed):
only looked at code because filip and Andres ran it and said it worked as expected

LeoWikstrom (dismissed):

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -141,6 +145,7 @@ void Renderer::DeleteDxResources()
 void Renderer::InitD3D12(const Window *window, HINSTANCE hInstance, ThreadPool* threadPool)
 {
 	m_pThreadPool = threadPool;
+	m_Window = window;
```
jocke1995: m_pWindow

LeoWikstrom (approved):

jocke1995 (approved):
heja heja!

PR #91, opened 2020-10-05
Dash

LeoWikstrom:


StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
```diff
@@ -87,6 +89,17 @@ void component::PlayerInputComponent::RenderUpdate(double dt)
...
+	std::chrono::duration<double> elapsed_time = std::chrono::system_clock::now() - m_DashTimer;
+	double time = elapsed_time.count();
+	m_DashReady = time > 1.5;
+	if (time > 0.5 && m_Dashing)
+	{
+		double3 vel = m_pCC->GetLinearVelocity();
+		vel /= 6.0;
+		m_pCC->SetVelVector(vel.x, vel.y, vel.z);
+		m_Dashing = false;
+	}
```
jakkeez: Wouldn't use of dt work as well, instead of external library?

jakkeez (commented):
Comment (Or question rather) about implementation, will probably approve if comment is solved (i.e answered)

StortSpelProjekt/Engine/src/ECS/Components/Collision/CapsuleCollisionComponent.cpp
```diff
@@ -14,5 +14,5 @@ component::CapsuleCollisionComponent::~CapsuleCollisionComponent()
 
 double component::CapsuleCollisionComponent::GetDistanceToBottom() const
 {
-	return dynamic_cast<btCapsuleShape*>(m_pShape)->getHalfHeight();
+	return dynamic_cast<btCapsuleShape*>(m_pShape)->getHalfHeight() + dynamic_cast<btCapsuleShape*>(m_pShape)->getRadius();
```
jocke1995: use static_cast, wrong usage of dynamic_cast

StortSpelProjekt/Vendor/config.txt
```diff
@@ -1,10 +1,10 @@
...
 f_backgroundVolume 0.2
 f_updateRate 30
-i_resolutionWidth 1920
-i_resolutionHeight 1080
-i_windowWidth 1920
-i_windowHeight 1080
+i_resolutionWidth 768
+i_resolutionHeight 432
+i_windowWidth 768
+i_windowHeight 432
```
jocke1995: Dont forget to change it back to 1920, 1080 when you merge

jocke1995 (dismissed):
Found two bugs when testing the dash:
1. Sometimes dashing out of the scene (tried several times and it doesn't happen everytime, so it might be hard to debug)
2. Dashing sideways (holding "w" and then double-tapping "a" or "d") sometimes dashes the wrong direction.

jakkeez (dismissed):
Minor fix is that in input I would be nice to utilize our implemented timer chronos instead of creating a new one. Nothing major, just a small fix that should be noted but can be fixed at a later point.

jocke1995 (dismissed):
Bra.

jakkeez (approved):

jocke1995 (approved):

PR #94, opened 2020-10-05
Connect by console

AntonAsbrink:
This features allows players to connect to the server by writing connect:[IP]
I also fixed so other players moves smoothly and have collision

To test this you need to start server.exe in the bin folder; write "1" followed by "AddClient" twice to add 2 slots. Then open up two instances of either sandbox or game. Using the console write "connect:localhost" on both instances and you will see another cube that is controlled by the other instance.

OBS! if you open up from the exe in bin. make sure to build it from VS first and move everything inside it to the corresponding folder by sln. the pathing to many files is not setup for /bin/****


StortSpelProjekt/Engine/src/ImGUI/ImGuiHandler.cpp
```diff
@@ -353,6 +353,13 @@ void ImGuiHandler::ExecCommand(const char* command_line)
     {
         m_BoolMap["reset"] = true;
     }
+    else if (Stricmp(command.c_str(), "CONNECT") == 0)
```
LeoWikstrom, line 356: Would be good to have a check if i_network in config is set to 1. If it is set to 0, the program crashes when typing the command

StortSpelProjekt/Engine/src/ImGUI/ImGuiHandler.cpp
```diff
@@ -353,6 +353,13 @@ void ImGuiHandler::ExecCommand(const char* command_line)
...
+    {
+        if (arguments.size() > 0)
+        {
+            EventBus::GetInstance().Publish(&ConnectToServer(arguments.at(0).c_str()));
```
LeoWikstrom, line 360: A default would be nice, though not required. If there is no argument, arguments[0] is -1. In this case it could perhaps be good to instead use s_ip from the config file?

StortSpelProjekt/Engine/src/ImGUI/ImGuiHandler.cpp
```diff
@@ -509,6 +516,7 @@ ImGuiHandler::ImGuiHandler()
     m_Commands.push_back("HISTORY");
     m_Commands.push_back("CLEAR");
     m_Commands.push_back("RESET");
+    m_Commands.push_back("CONNECT");
```
LeoWikstrom: Also here the check for i_network could be good, to keep the command from showing up in HELP and autocomplete when there is no network

StortSpelProjekt/Engine/src/Network/Network.cpp
```diff
@@ -23,10 +25,12 @@ bool Network::ConnectToIP(std::string ip, int port)
 
     if (status != 0) 
     {
+        ImGuiHandler::GetInstance().AddLog("Connection to %s failed", ip.c_str());
```
LeoWikstrom, line 28: After I connected successfully, I tried connecting to a random IP which of course failed. When I then tried to again connect to the correct IP, this message was printed, while the server said that I had been connected. However, no packets were sent. Is this expected behaviour?

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -72,21 +70,24 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int nCmdShow)
...
             physics->Update(updateRate);
         }
 
         /* ---- Network ---- */
-        if (networkOn)
+        if (network.IsConnected())
         {
-            if (networkCount == 2) {
-                networkCount = 0;
+            if (networkTimer >= networkUpdateRate) {
```
LeoWikstrom: Block should start on a new line, as per the code standard

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -290,7 +291,7 @@ Scene* GetDemoScene(SceneManager* sm)
     float xVal = 8;
     float zVal = 20;
     // extra 75 enemies, make sure to change number in for loop in DemoUpdateScene function if you change here
-    for (int i = 0; i < 75; i++)
+    for (int i = 0; i < 5; i++)
```
LeoWikstrom, line 295: I love it

LeoWikstrom (changes requested):
Looks sexy, just a few things to fix. Game works fine, but when running Sandbox the clients do not send any packets. Maybe look into that as well?

LeoWikstrom (dismissed):

jakkeez (dismissed):

jakkeez (approved):

LeoWikstrom (approved):

PR #96, opened 2020-10-07
Added connection logging in console on server-side and imgui client-side

jakkeez:
New logging on server-side in console and imgui on client-side when sending and receiving packages. Added a new debug flag as well to prevent this logging to clutter imgui and console if they are not interesting at the current time.


StortSpelProjekt/Server/src/ClientPool.cpp
```diff
@@ -141,16 +141,53 @@ void ClientPool::newPacket(int socket)
 	sf::Packet packet;
 	if (m_Clients.at(socket)->socket.receive(packet) == sf::Socket::Done)
 	{
-		m_ConsoleString.append("Recieved a packet from client " + std::to_string(socket) + "; " + std::to_string(packet.getDataSize()) + " BYTES\n");
```
AntonAsbrink, line 144: This should not be removed

AntonAsbrink (dismissed):

AntonAsbrink (approved):

LeoWikstrom (approved):
Everything seems to work

PR #98, opened 2020-10-07
Upgrade-component

kattalainen:
Created an UpgradeComponent, which all other specific upgradecomponents will have to inherit from. I made an example upgrade (UpgradeRangeTest) for testing purposes, which atm is attached to rangecomponent (for testing) and is triggered when projectile collides with the barbarian. AndresTestScene in Sandbox (UpgradeComponent-Ranged branch) is where you can test it out. It simply makes it so the projectile flies up in the sky on collision.


StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.cpp
```diff
@@ -0,0 +1,86 @@
+#include "UpgradeComponent.h"
+#include "stdafx.h"
+
+component::UpgradeComponent::UpgradeComponent(Entity* parent)
```
jocke1995, line 5: This class should not inherit from component. If it does, it means that it can be added by itself in main, which I dont think is the purpose of this component?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
...
+// Types of upgrades, for sorting/call purposes
+enum E_Type
+{
+	RANGE = 0,			// range weapon/attack upgrades
+	MELEE = 1,			// melee weapon/attack upgrades
+	MOVEMENT = 2,		// upgrades regarding movement.
+	STATS = 3,			// stat modifications such as hp change.
+	ONDAMAGE = 4,		// upgrades that are "triggered" when damage is taken.
+	ENEMYSPECIFIC = 5	// upgrades only relevant for enemy entities. example: enemy explode on death)
+};
```
jocke1995: If an upgrade should be able to have multiple types, you should use flags instead of normal enums. Also the name : E_Type is not explaining what sort of enum it is. Type of what?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace component
+{
+	class UpgradeComponent : public Component
```
jocke1995, line 10: Should not inherit from component

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
...
+		virtual ~UpgradeComponent();
+		virtual void Update(double dt);
+		virtual void RenderUpdate(double dt);
+
+		// setters and getters for private member variables
+		void SetName(std::string name);
+		std::string GetName();
+		void IncreaseLevel();
+		void DecreaseLevel();
+		int GetLevel();
```
jocke1995: const?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
...
+		void SetType(int type);
+		int GetType();
```
jocke1995: const?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
...
+	{
+	public:
+		UpgradeComponent(Entity* parent);
+		virtual ~UpgradeComponent();
+		virtual void Update(double dt);
+		virtual void RenderUpdate(double dt);
+
+		// setters and getters for private member variables
+		void SetName(std::string name);
+		std::string GetName();
```
jocke1995: const?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeComponent.h
```diff
@@ -0,0 +1,69 @@
+#ifndef UPGRADECOMPONENT_H
+#define UPGRADECOMPONENT_H
+#include "ECS/Components/Component.h"
+#include <vector>
```
jocke1995: why include vector? I dont see that you're using it in the headerfile?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/UpgradeRangeTest.cpp
```diff
@@ -0,0 +1,26 @@
...
+	IncreaseLevel();
+}
+
+component::UpgradeRangeTest::~UpgradeRangeTest()
+{
+}
+
+void component::UpgradeRangeTest::OnHit()
+{
+	Log::Print("UpgradeRangeTest OnHit called\n");
```
jocke1995: If this print is for testing, then its all good. But we should try to avoid to print stuff during runtime in release mode if its not really necessary.

jocke1995 (changes requested):
I haven't looked into your structure, meaning that how you planned to make this work in the future. I only ran your code and looked at your code.

kattalainen:
There is some restructuring going on, DO NOT APPROVE!

PR #99, opened 2020-10-08
Lock rotation

LeoWikstrom:


fredrix96 (approved):
Everything seems to be in order and the test scene is a piece of art

Schtekt (approved):
The Gandalf model should use PBR material in the future. The current material file is near useless.
The code is otherwise fine and the scene does not crash.
As Gandalf isn't used at the moment I will approve this :)

PR #104, opened 2020-10-08
Set origin of models to their center

LeoWikstrom:
Could not find a way to do it dynamically, must be done in blender. Fixed some other small stuff


jakkeez (approved):

Schtekt (approved):
Everything seems to be in order and the test scene is a piece of art

PR #109, opened 2020-10-12
FLAG_LIGHT::STATIC

jocke1995:


StortSpelProjekt/Engine/src/Renderer/HLSL/ForwardPixel.hlsl
```diff
@@ -59,28 +59,33 @@ PS_OUTPUT PS_main(VS_OUT input)
...
+				pointLight[index],
+				camPos,
+				viewDir,
+				input.worldPos,
+				metallic,
+				albedo.rgb,
+				roughness,
+				normal.rgb,
+				baseReflectivity);
+		//}
```
Whixx: Should this comment stay?

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -2104,8 +2111,8 @@ void Renderer::removeComponents(Entity* entity)
 
 void Renderer::prepareScenes(std::vector<Scene*>* scenes)
 {
-	prepareCBPerFrame();
-	prepareCBPerScene();
+	SubmitUploadPerFrameData();
```
Whixx, line 2110: Should be more comments to explain stuff like this. Why is the function called twice?

Whixx (changes requested):
Check comments, otherwise looks ok

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
Whixx: Nevermind, is not the same function :^)

PR #114, opened 2020-10-13
SetScene rework

fjz345:
Decicded on the daily meeting that new Scene* creates the resources and delete should remove them. Vram is still loaded when scene is set.


StortSpelProjekt/Engine/src/Misc/AssetLoader.h
```diff
@@ -50,6 +50,15 @@ class AssetLoader
...
 	std::pair<Font*, Texture*> LoadFontFromFile(const std::wstring& fontName);
 
+
+    // IsLoadedFunctions
+    bool IsModelLoadedOnGpu(const std::wstring& name) const;
+    bool IsModelLoadedOnGpu(const Model* model) const;
+    bool IsMaterialLoadedOnGpu(const std::wstring& name) const;
+    bool IsMaterialLoadedOnGpu(const Material* material) const;
+    bool IsTextureLoadedOnGpu(const std::wstring& name) const;
+    bool IsTextureLoadedOnGpu(const Texture* texture) const;
```
jocke1995, line 60: What do we need these functions for?

StortSpelProjekt/Engine/src/Misc/AssetLoader.h
```diff
@@ -65,13 +74,11 @@ class AssetLoader
...
-    bool IsModelLoadedOnGpu(const Model* model) const;
-    bool IsMaterialLoadedOnGpu(const std::wstring& name) const;
-    bool IsMaterialLoadedOnGpu(const Material* material) const;
-    bool IsTextureLoadedOnGpu(const std::wstring& name) const;
-    bool IsTextureLoadedOnGpu(const Texture* texture) const;
+    void loadDefaultMaterial();
+
+    // LoadVramFunctions
+    void loadMeshToGpu(Mesh* mesh);
+    void loadTextureToGpu(Texture* texture);
```
jocke1995: These functions doesn't have a body, did you forget to remove?

StortSpelProjekt/Engine/src/Renderer/Model.cpp
```diff
@@ -75,6 +75,22 @@ const SlotInfo* Model::GetSlotInfoAt(unsigned int index) const
...
+		{
+		m_Meshes[i]->m_pSRV->GetDescriptorHeapIndex(),
+		m_Materials[i]->GetTexture(TEXTURE2D_TYPE::ALBEDO)->GetDescriptorHeapIndex(),
+		m_Materials[i]->GetTexture(TEXTURE2D_TYPE::ROUGHNESS)->GetDescriptorHeapIndex(),
+		m_Materials[i]->GetTexture(TEXTURE2D_TYPE::METALLIC)->GetDescriptorHeapIndex(),
+		m_Materials[i]->GetTexture(TEXTURE2D_TYPE::NORMAL)->GetDescriptorHeapIndex(),
+		m_Materials[i]->GetTexture(TEXTURE2D_TYPE::EMISSIVE)->GetDescriptorHeapIndex()
+		};
+	}
+}
```
jocke1995: Is this meant to be called after a game-developer (for example) changes a texture on a model? If it isn't, isn't it better to just change the slotInfo automaticly when a new texture is set?

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -679,6 +681,53 @@ SwapChain* Renderer::GetSwapChain()
...
+		texture = model->GetMaterialAt(i)->GetTexture(TEXTURE2D_TYPE::EMISSIVE);
+		SubmitTextureToCodt(texture);
+	}
+}
+
+void Renderer::SubmitTextureToCodt(Texture* texture)
+{
+	CopyOnDemandTask* codt = static_cast<CopyOnDemandTask*>(m_CopyTasks[COPY_TASK_TYPE::COPY_ON_DEMAND]);
+	codt->SubmitTexture(texture);
+}
```
jocke1995: Very nice! :)

StortSpelProjekt/Engine/src/Renderer/Renderer.h
```diff
@@ -113,19 +115,10 @@ class Renderer
...
-	void UnloadModel(Model* model) const;
-	void UnloadMesh(Mesh* mesh) const;
-	void UnloadMaterial(Material* material) const;
-	void UnloadTexture(Texture* texture) const;
-
-	void UnloadRenderComponents();
+	void SubmitToCodt(std::tuple<Resource*, Resource*, const void*>* Upload_Default_Data);
+	void SubmitModelToCodt(Model* model);
+	void SubmitMeshToCodt(Mesh* mesh);
+	void SubmitTextureToCodt(Texture* texture);
```
jocke1995: If these are public, are they meant to be called by game-developer, or is it a misstake by you? If they are meant to be called by the game-developer. Then they would also need a function to execute the commandlist after data has been submitted.

StortSpelProjekt/Sandbox/src/main.cpp
```diff
@@ -123,6 +120,22 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int nCmdShow)
...
+            {
+               
+            }
+            else
+            {
+                
+            }
+            
+            sceneManager->SetScenes(1, activeScenes);
+        }
```
jocke1995: I modified this test to swap between my (jocke) and leos testScenes. When I spammed space, the program hung. I expect an issue to be WaitForGPU when you setScene.

jocke1995 (changes requested):
Crash when swapping fast between scenes. Probably issue with WaitForGPU. Also a few questions about other things.

StortSpelProjekt/Engine/src/Misc/AssetLoader.h
fjz345: Checks the pair<>.first value. Makes sure that stuff that is loaded does not get loaded. From my assumptions nothing in the game should currently get removed from memory but they can still be here for some usage.

StortSpelProjekt/Engine/src/Misc/AssetLoader.h
fjz345: Ops, removed-

StortSpelProjekt/Engine/src/Renderer/Model.cpp
fjz345: Do not know. Just thought this would be written at a couple of places in the code and it deserved a function. Don't know if it is something that the game-devs should be calling. I'll set it to private and call it on setModel/Mesh/Material

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
fjz345: k

StortSpelProjekt/Engine/src/Renderer/Renderer.h
fjz345: moved to private

jocke1995 (changes requested):
Remove test from main and then merge.

jocke1995 (approved):

LeoWikstrom (approved):

PR #116, opened 2020-10-14
ID:64 AddDynamic Entity

fjz345:
Important note: you have to make sure yourself that the model has been loaded if you add it to the scene.

The plan is that for example: ArrowShots
You in the scene/or straight in main call AssetLoader()->Load() on everything the arrow needs.

Then you just create new entity.


Schtekt (dismissed):
Not much to say. No problems found.

jocke1995 (changes requested):
Remove tests from main, then you're good to go

jocke1995 (dismissed):
You should've made a new bransch.

jocke1995 (approved):
Seems to work, finally

Schtekt (approved):

PR #119, opened 2020-10-14
Enemy Network

AntonAsbrink:
Enemies are updated based on the host, they enemies will attack the closest player


StortSpelProjekt/Engine/src/Network/Network.cpp
```diff
@@ -119,37 +119,23 @@ void Network::SendEnemiesPacket(std::vector<Entity*>* enemies)
     int nrOfEnemies
```
LeoWikstrom, line 119: Comments should be removed

StortSpelProjekt/Engine/src/Network/Network.cpp
```diff
@@ -317,6 +311,36 @@ void Network::processServerData(sf::Packet* packet)
     }
 }
 
+void Network::processEnemyData(sf::Packet* packet)
+{
+    /*
```
LeoWikstrom: Comments should be removed

StortSpelProjekt/Sandbox/src/main.cpp
```diff
@@ -28,6 +28,8 @@ void DemoUpdateScene(SceneManager* sm);
 
 void DefaultUpdateScene(SceneManager* sm);
 
+EnemyFactory enemyFactory;
```
LeoWikstrom, line 30: Should we maybe make each scene hold a pointer to their enemy factory? Maybe not in this task though...

StortSpelProjekt/Server/src/ClientPool.cpp
```diff
@@ -135,31 +137,47 @@ void ClientPool::enemyData(int index, sf::Packet packet)
 	int nrOfEnemies
```
LeoWikstrom, line 137: Remove comments

LeoWikstrom (changes requested):
Looks good, works well. Just a few comments to take care of

Schtekt (changes requested):
Nothing new to add from my part. Remove comments and merge in develop.

LeoWikstrom (approved):
🅰️

Schtekt (approved):
G

PR #126, opened 2020-10-16
SetSceneRemoveDynamic

fjz345:
Adding entities dynamic by creating entity in main with space and deleting them with tab. This code will be removed before pull.

GUI2DComponent is not done yet so code to remove it is not complete.

Since in the shader the number of lights is defined to a constant, it will crash if you space too much space if you have a light component added each time.


OtakuFreak13 (dismissed):
Runs fine. The abort that happens in some scenes is due to upgrade stuff so nothing wrong with this code.

fredrix96 (changes requested):
Code standard looks good but Filip wanted to fix something

OtakuFreak13 (approved):

fredrix96 (approved):
Seems fine

PR #127, opened 2020-10-19
Demoscene sprint 3

Schtekt:


jakkeez (dismissed):
No issues as I can see, only thing to note is the hack for the text. It is updated each frame instead of each hit, which tanks the FPS quite significantly. Fredrik has noted this and will fix at a later point.

kattalainen:
Update tutorial then it should be fine

kattalainen (approved):

jakkeez (approved):

PR #128, opened 2020-10-19
Added gravity per object instead of for the whole world.

Schtekt:


jakkeez (changes requested):
Error thrown in SetGravity()

jakkeez (approved):

OtakuFreak13 (approved):

PR #131, opened 2020-10-20
Create 2D GUIs, mark them and press them

fredrix96:
Make sure that the 2D GUIs are displayed at the screen and that Stefan's head is highlighted when you are moving your mouse to its position. When you left-click on the head it should log that is has been pressed


jakkeez (approved):

OtakuFreak13 (approved):
Clicking seems to work.

PR #134, opened 2020-10-21
ButtonPressed event

fredrix96:
When the button is pressed it publishes a ButtonPressed event which functions need to handle. Make sure that it says that the button has been pressed when you are pressing the head of Stefan


OtakuFreak13 (dismissed):
Works as intended in test scene and does not break game scene.

OtakuFreak13 (approved):
Still runs as intended.

Schtekt (approved):
Looks good!
am a bit annoyed over a "test" event, but it is passable.

PR #137, opened 2020-10-21
Smooth movement

Schtekt:
The player now "hovers" above the ground.
This changes how the player moves, affecting ALL entities using the playerinputcomponent.
A known issue is that regular shapes (boxes and spheres) fall through flat surfaces. We found this acceptable as these shapes are not to represent the player character any longer.


jakkeez (commented):
Currently experiencing problems with "Smooth Movement". After conditions are met (Tim knows that I am talking about), the player model falls right through the floor.

Schtekt:
The last error was caused by gravity being applied several times, as it was never "hitting" the ground the bt velocity automatically updated in the bullet physics black box.

Should work fine now 🥇

LeoWikstrom (dismissed):
Works well 👍

jakkeez (changes requested):
Sometimes when running the gravity goes bonkers and makes the player fly away.

Schtekt:
Gravity will now always kick in when player jumps, my bad!

jakkeez (approved):

LeoWikstrom (approved):

PR #138, opened 2020-10-21
Projectile Sound

jakkeez:
When firing a projectile, a sound "fireball"-sound now plays. However, currently, the sound is cut of if a new projectile is fired and rapid succession. This would not be an issue if there is a "cooldown" on the projectile firing. 
Currently looking for input on whether or not this is acceptable in its current state in regards to the current deadline. This can be revisited in the future.


kattalainen:
Code looks fine. I have manually edited the fireball sound file (cut of some of the beginning and end) and it is now barely even noticeable that the sound gets cut off when spamming projectiles and in my opinion this would totally be acceptable for this sprint. Will send it to you for testing.
The audio engine will definitely need some love in the future though!

Schtekt (dismissed):
9/10
A bit annoyed with the sound but it works as intended :)

kattalainen (approved):

Schtekt (approved):

PR #141, opened 2020-10-22
A new Game Level

AntonAsbrink:
A brand new level in game


Schtekt (dismissed):
Cute map, kind of small but I love it!

StortSpelProjekt/Engine/src/Misc/AssetLoader.cpp
```diff
@@ -12,6 +12,7 @@
 #include "../Renderer/Material.h"
 #include "../Renderer/Animation.h"
 #include "../Renderer/Transform.h"
+#include "../ECS/Components/SkyboxComponent.h"
```
kattalainen: I dont think you need to include skyboxcomponent here :>

kattalainen (commented):
Code is fine, but I have some questions about the map. Why isn't height map used? And in my opinion it should be a little bit more open because of the way our camera will end up behind the walls in big parts of the map. Let's talk about it tomorrow!

Schtekt (dismissed):

Schtekt (dismissed):

Schtekt (dismissed):

Schtekt (approved):

kattalainen (approved):

PR #147, opened 2020-10-22
Enemy attack interval setting

kattalainen:
Ai component now has an attack interval variable so rate of attacks can be regulated. Added argument for enemyfactory to set the interval.


StortSpelProjekt/Game/src/Gamefiles/Components/AiComponent.cpp
```diff
@@ -68,9 +70,20 @@ void component::AiComponent::Update(double dt)
...
+						Log::Print("ENEMY ATTACK!\n");
+						m_TimeAccumulator = 0.0;
+					}
 				}
 			}
+			else
+			{
+				// "reset" accumulator if out of attack range
+				m_TimeAccumulator = 0.0;
+			}
```
Schtekt, line 142: The cooldown should not be reset every time the enemy is out of range. It should be reset once the enemy has attacked.
For example, the enemy is outside attack range for a large amount of time, then attacks whenever the target is within range. There the cooldown starts instead.

fjz345 (dismissed):
Looks good.

Schtekt (changes requested):
I would like to see that "m_TimeAccumulator" only is reset after an attack has been made.
If for example, an enemy with a large cooldown (for a BIG attack) needs to wait while being inside attack range all the time, that would not be too good now, would it?

StortSpelProjekt/Game/src/Gamefiles/Components/AiComponent.cpp
kattalainen: I disagree, fight me

Schtekt (dismissed):

Schtekt (approved):

fjz345 (approved):

PR #148, opened 2020-10-23
Copy on demand fix

fjz345:
Copy on demand now is run every frame but often does nothing. This way we can submit meshes/textuers/etc anywhere in the code and then the GPU makes sure to load it in.

### DIFF
--- a/StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
+++ b/StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
@@ -345,6 +345,12 @@ void Renderer::Execute()
 	ComputeTask* computeTask = nullptr;
 	RenderTask* renderTask = nullptr;
 	/* --------------------- Record command lists --------------------- */
+
+	// Copy on demand
+	copyTask = m_CopyTasks[COPY_TASK_TYPE::COPY_ON_DEMAND];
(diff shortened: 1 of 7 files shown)


StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -2065,29 +2076,30 @@ void Renderer::waitForFrame(unsigned int framesToBeAhead)
...
+//}
+//
+//void Renderer::executeCopyOnDemand()
+//{
+//	m_CopyTasks[COPY_TASK_TYPE::COPY_ON_DEMAND]->SetCommandInterfaceIndex(0);
+//	m_CopyTasks[COPY_TASK_TYPE::COPY_ON_DEMAND]->Execute();
+//	m_CommandQueues[COMMAND_INTERFACE_TYPE::COPY_TYPE]->ExecuteCommandLists(1, &m_CopyOnDemandCmdList[0]);
+//	waitForCopyOnDemand();
+//	m_CopyTasks[COPY_TASK_TYPE::COPY_ON_DEMAND]->Clear();
+//}
```
fredrix96, line 2102: Should these lines of code still be here?

fredrix96 (approved):
Clever solution and it seems to work as intended. Good work!

LeoWikstrom (approved):

PR #149, opened 2020-10-23
Player Death Scene

fjz345:
There is a test that kills the player soon after start.


AntonAsbrink (changes requested):
Crashes when closing the window using ESC

AntonAsbrink (dismissed):

StortSpelProjekt/Engine/src/ECS/SceneManager.h
```diff
@@ -36,7 +42,12 @@ class SceneManager
 	std::vector<Scene*> m_ActiveScenes;
 	std::set<Scene*> m_LoadedScenes;
 
+	Scene* m_GameOverScene = nullptr;
```
fredrix96: You forgot a p for pointer

fredrix96 (changes requested):
Just a small code standard error, the rest looks to work well

fredrix96 (approved):
Seems to work as intended. Well done!

AntonAsbrink (approved):

PR #150, opened 2020-10-23
Triangle NavMesh support added

LeoWikstrom:
Support for triangle NavMesh has been added. SUpport for quad NavMesh is still working


kattalainen:
Code looks good and everything seems to be in order.
A little warning for next reviewer though: If you type in "boundingbox" in imgui to see the navmesh, the game will crash because "u" tries to apply upgrades on non-existing components. You can type in B then press tab to autocomplete the BOUNDINGBOX command instead :>

kattalainen (approved):

jakkeez (approved):
Nothing wrong as I can see, and code runs within the scenes.

PR #155, opened 2020-10-23
Tweak projectile cooldown

kattalainen:
Added cooldown for projectiles.


Schtekt (dismissed):
Good job!

StortSpelProjekt/Game/src/Gamefiles/Components/RangeComponent.cpp
```diff
@@ -56,6 +58,11 @@ void component::RangeComponent::OnUnInitScene()
 {
 }
 
+void component::RangeComponent::Update(double dt)
+{
+	m_TimeAccumulator += static_cast<float>(dt);
```
AntonAsbrink: I belive it is better to count down until you are allowed to fire again.
make m_TimeAccumulator to a double instead of casting to a float

StortSpelProjekt/Game/src/Gamefiles/Components/RangeComponent.cpp
```diff
@@ -66,61 +73,77 @@ float component::RangeComponent::GetVelocity()
...
-	tc->GetTransform()->SetScale(m_Scale);
-	tc->GetTransform()->SetVelocity(m_Velocity); 
-	tc->Update(0.02);
-	bbc = ent->AddComponent<component::BoundingBoxComponent>(F_OBBFlags::COLLISION);
-	bbc->Init();
-	Physics::GetInstance().AddCollisionEntity(ent);
-	if (audioPlay)
+void component::RangeComponent::Attack()
+{
+	if (m_TimeAccumulator > m_AttackInterval)
```
AntonAsbrink: minor detail but should be ">="
if we have we have update rate to 0.1 seconds, the true attack interval will be 0.6 seconds instead of the set 0.5

AntonAsbrink (changes requested):

AntonAsbrink (approved):

AntonAsbrink (approved):

Schtekt (approved):

PR #157, opened 2020-10-26
Enemy Health Tweak

LeoWikstrom:
Added function for easily changing max health of all enemies of a certain type. Difficult to set a finalised max hp, as we don't know exactly which kinds of enemy we will have, or how much damage the player will do


jocke1995 (dismissed):
Seems to work just fine! Remove Prints when changing hp and then merge.

kattalainen (dismissed):

kattalainen (dismissed):

kattalainen (dismissed):

OtakuFreak13 (approved):

kattalainen (approved):

PR #158, opened 2020-10-26
Removal of enemies on death.

Schtekt:
Removes enemies after a specified time after death.
Also added so that projectiles are removed after a certain time from creation.


jakkeez (changes requested):
Crash, plz fix bro

jakkeez (dismissed):

jakkeez (dismissed):

jakkeez (dismissed):

jakkeez (approved):
Good work. Whoever thought to implement two separate events for death and removal of entity is a genius !!

fredrix96 (approved):
Seems to work as intended

PR #160, opened 2020-10-26
Map props

AntonAsbrink:
Added some nice looking props to make the map a bit more busy


jocke1995 (dismissed):
Very good looking props!

Schtekt (approved):

jocke1995 (approved):

PR #163, opened 2020-10-26
Currency on Death

jakkeez:
When enemies die they now drop currency, the currency is specified by type and amount in currency component.

### DIFF
--- a/StortSpelProjekt/Game/src/main.cpp
+++ b/StortSpelProjekt/Game/src/main.cpp
@@ -6,9 +6,6 @@
 #include "Player.h"
 #include "Shop.h"
 
-// Game components includes
-#include "Gamefiles/Components/CurrencyComponent.h"
-
 Scene* GetDemoScene(SceneManager* sm);
 
 Scene* GameOverScene(SceneManager* sm);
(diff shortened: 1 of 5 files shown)


StortSpelProjekt/Game/src/main.cpp
```diff
@@ -6,6 +6,9 @@
 #include "Player.h"
 #include "Shop.h"
 
+// Game components includes
+#include "Gamefiles/Components/CurrencyComponent.h"
```
kattalainen: you can include this in GameComponents.h instead :>

StortSpelProjekt/Sandbox/src/main.cpp
```diff
@@ -3,6 +3,7 @@
 #include "GameNetwork.h"
 #include "Player.h"
 #include "Shop.h"
+#include "Components/CurrencyComponent.h"
```
kattalainen: GameComponents.h !

kattalainen (dismissed):
just move the currency includes to gamecomponent.h then it's fine

kattalainen (approved):

OtakuFreak13 (approved):
Runs and seems to add currency to player.

PR #168, opened 2020-10-27
tweaked melee damage

kattalainen:
Raised melee damage to 10. 
(Also increased enemy health to 30 so they wont get one-shot. Mass enemy health tweak "between levels" should then make use of Leos "enemy health tweak" functions, enemyFactory.SetEnemyTypeMaxHealth())


LeoWikstrom (approved):

fredrix96 (approved):
Good work!

PR #171, opened 2020-10-28
Player health tweak

kattalainen:
Lowered player hp to 50. With enemy attacks at 10 the player can take 5 hits before death. Might seem kind of low but keep in mind there are three upgrades that either gives more hp, life steal or lower damage taken when under 30% hp. Start low and work your way up!


LeoWikstrom (approved):
I like how we're making Melee attacks less and less viable 👍

kattalainen:
> 
> 
> I like how we're making Melee attacks less and less viable 👍

There's melee attacks in this game!?

OtakuFreak13 (approved):

PR #180, opened 2020-11-02
Only 1 scene active, uninit runs.

fjz345:


jakkeez (dismissed):

StortSpelProjekt/Engine/src/ECS/SceneManager.h
```diff
@@ -45,6 +46,7 @@ class SceneManager
 	SceneManager();
 	std::map<std::string, Scene*> m_Scenes;
 	std::vector<Scene*> m_ActiveScenes;
+	Scene* m_DefaultScene;
```
fredrix96: Code standard. It should be m_pDefaultScene

fredrix96 (changes requested):

StortSpelProjekt/Engine/src/ECS/SceneManager.h
fjz345: fixed

fredrix96 (approved):
Works fine

jakkeez (approved):

PR #182, opened 2020-11-02
Level kill goal

AntonAsbrink:
This feature adds a goal to kill X amount of enemies before the portal spawns


Schtekt (changes requested):
When the player dies the bar goes up. Is it possible to avoid this? otherwise I love it!

OtakuFreak13 (dismissed):
Code looks good. Tested 3 "levels" and worked as intended. 
We now have game loop nice!

OtakuFreak13 (approved):

Schtekt (approved):
All good

PR #183, opened 2020-11-02
It is now possible to update texts and quads during runtime

fredrix96:
The text and quad is updated in FredriksTestScene

### DIFF
--- a/StortSpelProjekt/Engine/src/ECS/Components/GUI2DComponent.h
+++ b/StortSpelProjekt/Engine/src/ECS/Components/GUI2DComponent.h
@@ -24,8 +24,8 @@ namespace component
 	private:
 		friend class Renderer;
 
-		TextManager m_pTextMan;
-		QuadManager m_pQuadMan;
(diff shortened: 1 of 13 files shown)


StortSpelProjekt/Engine/src/ECS/Components/GUI2DComponent.h
```diff
@@ -32,11 +26,6 @@ namespace component
 
 		TextManager m_pTextMan;
 		QuadManager m_pQuadMan;
```
OtakuFreak13: Only pointers should have _p as part of their name

OtakuFreak13 (changes requested):

Schtekt (changes requested):
Shop scene in sandbox crashes. Seems as there are no entities when it searches for a specific entity.
Looks like it was removed in develop, so try merging with develop first.
Everything else seems fine though :D

OtakuFreak13 (approved):

Schtekt (approved):

PR #185, opened 2020-11-02
Git ignore for range enemy model

OtakuFreak13:
Now all models that should be ignored are kept in the new folder "IgnoredModels".

Still retaining the old ignores since we can't have any mistakes happening when people upload stuff from older branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,4 @@ StortSpelProjekt/Vendor/Resources/Models/Female/
 
 StortSpelProjekt/Vendor/Resources/Models/PlayerHuman/
 
-StortSpelProjekt/Vendor/Resources/Models/Demon/
-
 StortSpelProjekt/Vendor/Resources/Models/IgnoredModels/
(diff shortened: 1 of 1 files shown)


Schtekt (approved):

.gitignore
```diff
@@ -23,3 +23,7 @@ StortSpelProjekt/Game/imgui.ini
 StortSpelProjekt/Vendor/Resources/Models/Female/
 
 StortSpelProjekt/Vendor/Resources/Models/PlayerHuman/
+
+StortSpelProjekt/Vendor/Resources/Models/Demon/
```
Schtekt: If you want to have one folder for all ignores I would like to ask that you remove these ignores from the gitignore.

Schtekt (changes requested):

.gitignore
OtakuFreak13: Yeah I'm the only one that has used it so it should be removed. 

Still keeping the rest though to avoid people forgetting to move over older models and accedentaly uploading stuff.

Schtekt (approved):

fredrix96 (approved):
Not much to say

PR #186, opened 2020-11-03
CollisionCategorisation.

Schtekt:
As a test of collisioncategories the player and projectile do not send collisionevent anymore. Instead they do a print.

This solution has flaws, scaling with many categories will be difficult and it all needs to be defined in the engine project.


AntonAsbrink (approved):

Whixx (approved):
well done

PR #191, opened 2020-11-04
We now have a main menu!

Schtekt:
The game scene now displays a menu where the user may start the game or exit with mouseclicks.


StortSpelProjekt/Game/src/Gamefiles/MainMenuHandler.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    Scene* scene = sm->CreateScene("MainMenuScene");
+
+    component::GUI2DComponent* guic = nullptr;
+
+    Texture* startTex = al->LoadTexture2D(L"../Vendor/Resources/Textures/2DGUI/Start.png");
+    Texture* optionsTex = al->LoadTexture2D(L"../Vendor/Resources/Textures/2DGUI/Options.png");
+    Texture* exitTex = al->LoadTexture2D(L"../Vendor/Resources/Textures/2DGUI/Exit.png");
+
+    // Player (Need a camera)
```
LeoWikstrom, line 41: Love it ❤️

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -609,6 +610,52 @@ Scene* ShopScene(SceneManager* sm)
     return scene;
 }
 
+void logSchtuff(const std::string& name)
```
LeoWikstrom: What is this?

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -609,6 +610,52 @@ Scene* ShopScene(SceneManager* sm)
...
+{
+    Log::Print("%s\n", name.c_str());
+}
+
+Scene* MainMenuScene(SceneManager* sm)
```
LeoWikstrom: Why is this scene created in two places?

LeoWikstrom (changes requested):
Make the mouse be released as default while in menu

StortSpelProjekt/Game/src/main.cpp
Schtekt: Something that should not exist, thank you for spotting it 👍

StortSpelProjekt/Game/src/main.cpp
Schtekt: it was a copy pasta from main instead of cut and pasta. The function is removed from main.

LeoWikstrom (changes requested):
Menu buttons can be clicked while in game, even though not visible

LeoWikstrom (dismissed):
It is approved

jocke1995 (dismissed):
so good, yes so perfect! yes yes yes!

StortSpelProjekt/Game/src/Gamefiles/Components/TeleportComponent.cpp
```diff
@@ -38,6 +38,7 @@ void component::TeleportComponent::OnCollision(Collision* collisionEvent)
 		else
 		{
 			Player::GetInstance().IsInShop(false);
+			EventBus::GetInstance().Publish(&RoundStart());
```
jocke1995, line 41: Jacob told me a few weeks back that its usually bad practise to publish events inside events, did you think this structure through before you did this?

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -52,11 +53,10 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int nCmdShow)
     Scene* demoScene = GameScene(sceneManager);
     Scene* shopScene = ShopScene(sceneManager);
     Scene* gameOverScene = GameOverScene(sceneManager);
+    Scene* mainMenuScene = MainMenuHandler::GetInstance().CreateScene(sceneManager);
 
-    //Scene* shopScene = ShopScene(sceneManager);
-    sceneManager->SetScene(demoScene);
+    sceneManager->SetScenes(mainMenuScene);
```
jocke1995: Its not called SetScenes any more. its called SetScene. This gave me compilation error.

StortSpelProjekt/Engine/src/ECS/SceneManager.cpp
```diff
@@ -116,7 +116,7 @@ void SceneManager::ChangeScene()
 {
 	if (m_ChangeSceneNextFrame)
 	{
-		if (m_pActiveScene->GetName() == "ShopScene" || m_pActiveScene->GetName() == "GameScene")
+		if (m_ActiveScenes[0]->GetName() != "gameOverScene")
```
jocke1995: this should be m_pActiveScene, also gave me a compilation error

jocke1995 (dismissed):
Almost! Just a few compilation errors

StortSpelProjekt/Game/src/main.cpp
Schtekt: Fixed

StortSpelProjekt/Engine/src/ECS/SceneManager.cpp
Schtekt: Fixed

StortSpelProjekt/Game/src/Gamefiles/Components/TeleportComponent.cpp
Schtekt: The only other option I see to this is to add an update to the teleportcomponent that checks if a boolean is true or not to call the roundstart event. For that reason I don't see another solution to this problem. I would like to know why this is bad practice though?

LeoWikstrom (approved):

OtakuFreak13 (approved):

PR #194, opened 2020-11-04
Currency gui

AntonAsbrink:
The gui now shows up in shop too, with a nice little icon.

I added independat gui entites for each scene which i belive is sensible since we can edit the gui by scene basis.


jocke1995 (changes requested):
The art of the currency png is ugly. If the currency texture is meant to be "temporary" just to make it work until willes TA-friend creates real textures, then it's fine. But if this is meant to be the final currency texture, then I vote big nono

AntonAsbrink:
I'm fine with any changes to the art anyone makes. But i will not change it!
I'm not gonna spend time on making nice art as i'm not an artist. Either accept the pullrequest and it's art as temporary and slated for change when avaible or we close this request and let someone else do it

jocke1995 (approved):

jakkeez (approved):

PR #198, opened 2020-11-05
Camera Distance Lock

jakkeez:
The camera now respects boundaries and "zooms in" when it hits a wall or floor. 
The camera distance can be changed in playerinput header file, in a define.
Test by going into game and moving around, turning the camera and walking up to the wall to see if it clips, if it doesnt, the implementation works.


StortSpelProjekt/Game/src/Gamefiles/Components/MeleeComponent.h
```diff
@@ -39,12 +39,12 @@ namespace component
 		void Update(double dt);
 		void Attack();
 		
-		void setAttackInterval(float interval);
+		void SetAttackInterval(float interval);
```
jocke1995, line 42: I think this change made something in main.cpp not to compile. In main you still write this with lowerCase S. I had to comment out that line to compile the code

jocke1995 (changes requested):
1. Didn't compile in main because of new name of setAttackIntervall.
2. Talk to me (Joakim), I found some bugs with the camera clipping that we can discuss when you get back to solve this 👍

Whixx (commented):
Camera sensitivity "goes up" when the camera zooms.

Whixx (dismissed):
Good work!

jocke1995 (dismissed):
Works, just 1 small clipping which we discussed when I reviewed this. But I think its fine! gj

Whixx (approved):

jocke1995 (approved):
works good

PR #199, opened 2020-11-05
Play/Pause Animations

Whixx:
In my testScene, check so that the animation is displayed while holding W and paused when you release W. The animation will be turned towards the player, which is wrong but is caused by the model. When we implement a model that is not just for testing i can add the relevant animations for when the player does the corresponding action and also reset the animation so that the new animation starts from keyframe 0.


StortSpelProjekt/Engine/src/Renderer/AnimatedModel.h
```diff
@@ -31,11 +31,15 @@ class AnimatedModel : public Model
     const ConstantBuffer* GetConstantBuffer() const;
     const std::vector<DirectX::XMFLOAT4X4>* GetUploadMatrices() const;
 
+    bool SetActiveAnimation(std::string animationName);
     void Update(double dt);
+    void toggleAnimation();
```
fredrix96: code standard

StortSpelProjekt/Engine/src/Renderer/AnimatedModel.h
```diff
@@ -31,11 +31,15 @@ class AnimatedModel : public Model
...
 
 private:
+    void initializeAnimation(SkeletonNode* node);   // Attach the currentStateTransforms of the animation to the skeleton.
     void updateSkeleton(float animationTime, SkeletonNode* node, DirectX::XMMATRIX parentTransform);
 
+    bool animationIsPaused = true;
```
fredrix96: code standard

StortSpelProjekt/Engine/src/Renderer/AnimatedModel.h
jocke1995: t should be capital letter

StortSpelProjekt/Engine/src/Renderer/Animation.h
```diff
@@ -12,8 +12,11 @@ struct BoneInfo
 struct TransformKey
 {
 	DirectX::XMFLOAT3 position = {};
+	DirectX::XMFLOAT3* pPosition = nullptr;
 	DirectX::XMFLOAT4 rotationQuaternion = {};
-	DirectX::XMFLOAT3 scaling = {};
+	DirectX::XMFLOAT4* pRotation = nullptr;
+	DirectX::XMFLOAT3 scaling = { 1.0f,1.0f,1.0f };
+	DirectX::XMFLOAT3* pScale = nullptr;
```
jocke1995, line 19: code standard with variable names

StortSpelProjekt/Engine/src/ECS/Components/AnimationComponent.h
```diff
@@ -0,0 +1,30 @@
...
+        void OnInitScene();
+        void OnUnInitScene();
+
+        void Initialize();
+
+        void SetActiveAnimation(std::string animationName);
+
+    private:
+        void walkAnimation(MovementInput* evnt);
+        AnimatedModel* m_pModel;
```
jocke1995: m_pAnimatedModel

StortSpelProjekt/Engine/src/ECS/Components/ModelComponent.cpp
```diff
@@ -31,11 +31,6 @@ namespace component
 
 	void ModelComponent::Update(double dt)
 	{
-		AnimatedModel* animatedModel = dynamic_cast<AnimatedModel*>(m_pModel);
-		if (animatedModel)
-		{
-			animatedModel->Update(dt);
-		}
```
jocke1995, line 38: Are you sure about this?  Did ModelComponent not have a update before?

fredrix96 (dismissed):
Would be good if you could walk to the sides as well

jocke1995 (changes requested):
Some code changes. Also some thoughts on how this works. But since we have talked about this, its your own choice if you want to fix these issues, or if you think that it isn't possible to fix it before we get an object with the animations that we are supposed to use in the game.
1. Doesn't animate when walking backwards / sideways.
2. Animation "freezes" in the middle of the animation when I stopped pressing w, and then returned by pressing w again. Instead the animation should reset when w is pressed again. Later on we will probably have an idle animation that will run when w isn't pressed, so then it would look wierd if the walk animation would start "In the middle of the animation sequence".
3. The model is rotated the opposite way that it is supposed to. (I dont know the reason for this, but it should be fixed)

jocke1995 (dismissed):
1. Crashes half of the times when starting.
2. Sometimes randomly falls through the floor in your branch (WIlliamsTestScene)

StortSpelProjekt/Engine/src/Renderer/AnimatedModel.h
LeoWikstrom: Same. Would also prefer to have the initialisation in the constructor

StortSpelProjekt/Engine/src/Renderer/Animation.h
LeoWikstrom: We have not discussed code standard for variables in structs. To me it looks okay

StortSpelProjekt/Sandbox/src/main.cpp
```diff
@@ -297,10 +298,13 @@ Scene* LeosTestScene(SceneManager* sm)
     melc = entity->AddComponent<component::MeleeComponent>();
     ranc = entity->AddComponent<component::RangeComponent>(sm, scene, sphereModel, 0.3, 1, 20);
     uc = entity->AddComponent<component::UpgradeComponent>();
+    ac = entity->AddComponent<component::AnimationComponent>();
 
 
     mc->SetModel(playerModel);
-    mc->SetDrawFlag(FLAG_DRAW::DRAW_ANIMATED | FLAG_DRAW::GIVE_SHADOW);
+    mc->SetDrawFlag(FLAG_DRAW::DRAW_ANIMATED | FLAG_DRAW::GIVE_SHADOW | FLAG_DRAW::NO_DEPTH);
```
LeoWikstrom, line 305: Why should the player have no depth flag?

LeoWikstrom (dismissed):
Just code standard things. Works well 👍

StortSpelProjekt/Sandbox/src/main.cpp
Whixx: Because the animations are applied in a separate depth pass. This means the regular depth pass should not run, else we basically run depth twice for the animated models.

LeoWikstrom (approved):

fredrix96 (approved):
Looks like a well-done work. It is nice to see working animations!

PR #200, opened 2020-11-06
Camera/movement fixes

LeoWikstrom:
Player can now move around freely. When camera rotates, player rotates as well. If attacking, player turns in camera direction. Player turns back to normal after 1.5 attack intervals, or when pressing sprint or dash.


jocke1995 (approved):
Everything seems to work.

jakkeez (approved):

PR #201, opened 2020-11-06
Dynamic Entity Fix

jocke1995:
Lights will now not spawn from origo when shooting.
Also, if you shot just before you change scene, lights will now get deleted and not be visible in the new scene as it was before (so did a bugfix) :))


fredrix96 (approved):
The problem is gone, good job

OtakuFreak13 (approved):

PR #204, opened 2020-11-09
CameraStutter

LeoWikstrom:
Made camera stutter more bearable and fixed assertion error


StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
```diff
@@ -8,6 +8,8 @@
 #include "Physics/Physics.h"
 #include "../Misc/Option.h"
 
+#include "stdafx.h"
+
```
Schtekt, line 12: Feels a bit weird that you'd use the pch in Game project. But after some reading on the matter I can accept it. Make sure that this is the first include though!

Schtekt (approved):
I love it!

AntonAsbrink (approved):

PR #207, opened 2020-11-10
Options Menu

AntonAsbrink:
The options menu changes values in the option class and consequently almost all changes require an restart to take effect. I implore anyone to make the changes dynamic if able but otherwise this is what it is.


StortSpelProjekt/Engine/src/Misc/Window.cpp
```diff
@@ -225,12 +225,12 @@ const HWND* Window::GetHwnd() const
 	return &m_Hwnd;
 }
 
-void Window::SetScreenWidth(int width)
+void Window::SetScreenWidth(const int width)
```
LeoWikstrom: Why const? It's not a pointer or a reference, so the source can't be changed anyway..

StortSpelProjekt/Game/src/Gamefiles/GameGUI.cpp
```diff
@@ -33,6 +34,12 @@ void GameGUI::Update(double dt, Scene* scene)
 		}
 	}
 
+	if (scene->GetName() == "OptionScene")
+	{
+		Entity* entity = scene->GetEntity("volume");
+		entity->GetComponent<component::GUI2DComponent>()->GetTextManager()->SetText(Option::GetInstance().GetVariable("f_volume"), "volume");
```
LeoWikstrom, line 40: I suppose this is to make the text "volume" appear? It does not show up for me
image

LeoWikstrom (changes requested):
The text "volume" does not show up. It is also possible to set a negative volume, this should be capped to 0, should it not? In my mind it would also be more intuitive to have the + on the right and the - on the left, but that's a personal preference I guess. And finally I assume there was no way to make the green quads move? It looks a bit odd when you select an option and it seems as if nothing happens

StortSpelProjekt/Engine/src/Misc/Window.cpp
AntonAsbrink: stuff left from trying to make it work dynamically. There was a lot of shenanigans concerning const and windows.

Whixx (changes requested):
Same stuff as Leo said

StortSpelProjekt/Game/src/Gamefiles/GameGUI.cpp
AntonAsbrink: There is no text marking that as volume, i was just so used to it being volume i didn't question it

LeoWikstrom (dismissed):
I'd still prefer the + to the right, but everything works. Good job 👍🏽

Whixx (dismissed):
I don't really like that we need this many different textures saved in the project for this to work, but i guess it's fine for now.

LeoWikstrom (approved):

Whixx (approved):
brao

PR #208, opened 2020-11-10
You can now aim towards a crosshair

fredrix96:


OtakuFreak13 (approved):

kattalainen (commented):
Gets kinda inaccurate when velocity is at 100 and when the camera is "zoomed in" because of walls. Lets sort it out tomorrow

PR #209, opened 2020-11-10
Added "Potion of strength" upgrade

Schtekt:
LÅG PRIO, INTE DEL AV SPRINT


StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradePotionOfStrength.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+}
+
+void UpgradePotionOfStrength::IncreaseLevel()
+{
+	m_Price = (++m_Level + 2) * 100;
+	ApplyStat();
+}
+
+void UpgradePotionOfStrength::ApplyStat()
```
OtakuFreak13, line 31: If you want it to take effect the first time you buy it then you have to use the function ApplyBoughtUpgrade() and call on ApplyStat() in there. This should probably be said in the tutorial my bad.

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradePotionOfStrength.cpp
```diff
@@ -0,0 +1,66 @@
+#include "UpgradePotionOfStrength.h"
+#include "EngineMath.h"
+#include "ECS/Entity.h"
+
+UpgradePotionOfStrength::UpgradePotionOfStrength(Entity* parent) : Upgrade(parent)
+{
+	// Set the name of the upgrade
+	SetName("PotionOfStrength");
```
OtakuFreak13, line 8: Please add "Upgrade" to the begining of the name so it follows the naming convention of the other upgrades. Needed since this is built upon by other parts of our game such as UpgradeGUI.

OtakuFreak13 (changes requested):

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradePotionOfStrength.cpp
OtakuFreak13: Now that I look through the shop and upgradeManager code it seems like the increase in level for UpgradeManager is actually applying your damage change. This is a bug that I need to fix since it clashes, in other upgrades, with ApplyBoughtUpgrade(). I need to take a closer look at this.

OtakuFreak13: This bug is fixed in my UpgradeGUI pullrequest. So add the ApplyBoughtUpgrade() and call ApplyStat() in it.

PR #210, opened 2020-11-10
Upgrade GUI menu

OtakuFreak13:
Hit "u" to open up the menu. 
When an upgrade is bought you can click on the upgrade "buttons"  that appear in the menu to show their description.
This menu will NOT update if you buy an upgrade while the menu is open.
Just close and open the menu again to reload.

To check the function for cycling through all upgrades, if you have more than 10 upgrades. You have to uncomment  the map called vec and change which forloop is commented out  in update() and createMenu().


StortSpelProjekt/Game/src/Gamefiles/UpgradeGUI.cpp
```diff
@@ -0,0 +1,510 @@
...
+			// Delete existing buttons.
+			for (int i = 0; i < m_ButtonNames.size(); i++)
+			{
+				if (m_CurrentScene->EntityExists(m_ButtonNames[i]))
+				{
+					m_Sm->RemoveEntity(m_CurrentScene->GetEntity(m_ButtonNames[i]), m_CurrentScene);
+				}
+			}
+
+			int itterator = 0;
```
LeoWikstrom: Iterator is spelled with one 't' 😘

StortSpelProjekt/Game/src/Gamefiles/UpgradeGUI.cpp
```diff
@@ -0,0 +1,510 @@
...
+			int posItterator = 0;
```
LeoWikstrom: Same here 😁

StortSpelProjekt/Game/src/Gamefiles/UpgradeGUI.cpp
```diff
@@ -0,0 +1,510 @@
...
+		blended,
+		m_boardBackgroundTexture, {0.4, 0.4, 0.4});
+	m_Sm->AddEntity(entity, scene);
+	entity->Update(0);
+	entity->SetEntityState(true);
+
+	/* ------------------------- Upgrade Menu Background End --------------------------- */
+
+	/* ------------------------- Upgrade Menu Buttons --------------------------- */
+	int itterator = 0;
```
LeoWikstrom: One 't'

StortSpelProjekt/Game/src/Gamefiles/UpgradeGUI.h
```diff
@@ -0,0 +1,73 @@
...
+	// How many times the menu can be filled with buttons.
+	// Used for keeping track of how many times we need to cycle
+	// the menu to show all buttons.
+	int m_ButtonsMultipleTen = 0;
+	// How many times we have filled the menu with buttons.
+	// Used for cycling through buttons.
+	int m_TimesFilledMenu = 0;
+	std::string m_CurrentDescription = "";
+	Scene* m_CurrentScene = nullptr;
+	SceneManager* m_Sm = nullptr;
```
LeoWikstrom: Pointers. Should be called m_pCurrentScene and m_pSm

StortSpelProjekt/Game/src/Gamefiles/UpgradeGUI.h
```diff
@@ -0,0 +1,73 @@
...
+
+	// Textures:
+	Texture* m_deviderTexture;
+	Texture* m_buttonMintTexture;
+	Texture* m_buttonElipseTexture;
+	Texture* m_orangeBackgroundTexture;
+	Texture* m_yellowGradientTexture;
+	Texture* m_orangeGradientTexture;
+	Texture* m_boardBackgroundTexture;
+	Texture* m_buttonParchment;
```
LeoWikstrom: Capital letter after m_
Also, divider is spelled with an 'i'

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -235,7 +240,7 @@ Scene* GameScene(SceneManager* sm)
 
 	EnemyComps zombie = {};
 	zombie.model = enemyModel;
-	zombie.hp = 30;
+	zombie.hp = 1;
```
LeoWikstrom: Don't forget to change this back

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -422,7 +427,7 @@ Scene* GameScene(SceneManager* sm)
 		killedEnemiesHolderTexture
 	);
 
-    /* --------------------------- GUI ------------------------------- */
+    /* ------------------------ GUI END ---------------------------- */
```
LeoWikstrom, line 454: Thank you for adding the "END"

LeoWikstrom (changes requested):
Works well overall. The HealthBoost actually seems to give 20 extra hp, and maybe it should be mentioned that it's 20 per level? But maybe that is not really part of this task 😅 It would, however, be nice to see the level of the upgrade as well

OtakuFreak13:
> 
> 
> Works well overall. The HealthBoost actually seems to give 20 extra hp, and maybe it should be mentioned that it's 20 per level? But maybe that is not really part of this task 😅 It would, however, be nice to see the level of the upgrade as well

This is a bug that I just found when reviewing Tims "Potion of strength" upgrade, the level of an upgrade is increased when an upgrade is bought in the shop. This then leads to the upgrade being applied twice since it is applied and then increased in level which then applies it again. Just pushed a fix for this.

Yeah I should definitely add the level of the upgrade to the description.

kattalainen:
Like Leo said it would be nice to see what level the upgrades are :> 
Besides this I have also noticed a bug in the game scene where I get sent back to the main menu for some reason, I will try to find a way of reproducing the bug.

LeoWikstrom (dismissed):

kattalainen (approved):

LeoWikstrom (approved):

PR #211, opened 2020-11-10
Ranged enemy can be managed in enemy factory

kattalainen:
Made a ranged enemy define in gamescene. The enemyfactory will spawn each fifth enemy as a ranged demon. "Testing" done in game and not in any sandbox scene.


OtakuFreak13 (dismissed):
Seems to work. Spawns in a range enemy for every 5th enemy spawned which seems to be correct.

LeoWikstrom (changes requested):
Seems to work, however a specific currency amount for the enemy type needs to be added in CurrencyComponent.cpp

LeoWikstrom (approved):

OtakuFreak13 (approved):

PR #212, opened 2020-11-11
OnePixelSmaller

LeoWikstrom:
Resolution width is now set to one pixel smaller than the screen size if equal to the screen size


OtakuFreak13 (approved):

fredrix96 (approved):
Goodbye darkness my old friend

PR #214, opened 2020-11-11
Barrels do not glide forever + basic knockback mah dudes

LeoWikstrom:


OtakuFreak13 (dismissed):
Works fine. Noticed a barrel that was shot flying through the wall, but couldn't reproduce it so think it's just the same thing as when we can dash through walls so not a bug in this task.

AntonAsbrink (changes requested):
projectile gets different speeds, probebly hitdir vector is not normalized

OtakuFreak13 (dismissed):

AntonAsbrink (dismissed):

AntonAsbrink (approved):

OtakuFreak13 (approved):
Seems to work.

PR #219, opened 2020-11-12
Added new sound when enemies are hit

jakkeez:
The "bruh" sound has been replaced with a demon sound for all enemies. The sound "Demon_Vocalisation_GotHit_1"  will have to be acquired and placed in the audiofolder and named DemonGotHit1 for the sound to work


LeoWikstrom (dismissed):

LeoWikstrom (approved):

kattalainen (commented):
Sound is fine, but it is kinda harsh on the ears when hitting multiple targets :) Might be better to wait to merge this until the sound amplification gets fixed or melee attack changed.

jakkeez:
Will be fixed at a later point when sounds do not amp each other.

PR #223, opened 2020-11-12
CopyOnDemand fix

fredrix96:
Many of the copyOnDemand related crashes should now be gone


StortSpelProjekt/Engine/src/Renderer/DX12Tasks/CopyOnDemandTask.cpp
```diff
@@ -31,6 +31,56 @@ void CopyOnDemandTask::SubmitTexture(Texture* texture)
 	m_Textures.push_back(texture);
 }
 
+void CopyOnDemandTask::UnSubmitMesh(Mesh* mesh)
+{
+	// Erase uploaded data
+	Resource* uploadResourceVertices = mesh->m_pUploadResourceVertices;
```
LeoWikstrom, line 37: Should not both of these be "Default" instead of "Upload" or have I misunderstood?

StortSpelProjekt/Engine/src/Renderer/DX12Tasks/CopyOnDemandTask.cpp
```diff
@@ -31,6 +31,56 @@ void CopyOnDemandTask::SubmitTexture(Texture* texture)
...
+			break;
+		}
+		it++;
+	}
+}
+
+void CopyOnDemandTask::UnSubmitText(Text* text)
+{
+	// Erase uploaded data
+	Resource* uploadResource = text->m_pUploadResourceVertices;
```
LeoWikstrom, line 68: Same as above. Is this correct or should it be "Default"?

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
```diff
@@ -714,16 +714,20 @@ void Renderer::InitBoundingBoxComponent(component::BoundingBoxComponent* compone
 	{
 		for (unsigned int i = 0; i < component->GetNumBoundingBoxes(); i++)
 		{
-			Mesh* m = BoundingBoxPool::Get()->CreateBoundingBoxMesh(component->GetPathOfModel(i));
-			if (m == nullptr)
+			auto[mesh, toBeSubmitted] = BoundingBoxPool::Get()->CreateBoundingBoxMesh(component->GetPathOfModel(i));
```
LeoWikstrom, line 717: This line is sexy as fuck

LeoWikstrom (changes requested):
It's difficult to test if it actually works as intended. Any tips for a specific scenario to test, or is it very sporadic and random?

jocke1995:
> It's difficult to test if it actually works as intended. Any tips for a specific scenario to test, or is it very sporadic and random?

It is very sporadic. If you play the game for about 3-4 rounds, it crashed most of the times before this fix.

StortSpelProjekt/Engine/src/Renderer/Renderer.cpp
jocke1995: c++17 boii

StortSpelProjekt/Engine/src/Renderer/DX12Tasks/CopyOnDemandTask.cpp
jocke1995: It is correct. In this case it doesn't matter. This function is only using the == operator to find the mesh to be unsubmitted. It could be used with defaultResource aswell. Since a mesh have its unique upload/default resource anyways

StortSpelProjekt/Engine/src/Renderer/DX12Tasks/CopyOnDemandTask.cpp
jocke1995: see above

LeoWikstrom (approved):

Whixx (approved):

PR #224, opened 2020-11-12
BouncyProjectiles

LeoWikstrom:


StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeBounce.h
```diff
@@ -0,0 +1,20 @@
...
+
+	void IncreaseLevel();
+	void ApplyBoughtUpgrade();
+	std::string GetDescription(unsigned int level);
+
+private:
+
+};
+
+#endif // !UPGRADE_BLUE_JEWEL_H
```
kattalainen: copy paste from blue jewel upgrade :>

StortSpelProjekt/Game/src/Gamefiles/Components/RangeComponent.cpp
```diff
@@ -91,6 +92,11 @@ float component::RangeComponent::GetAttackInterval() const
 	return m_AttackInterval;
 }
 
+void component::RangeComponent::MakeProjectileBouncy(bool bouncy)
+{
+	m_ProjectileRestitution = 0.5f * static_cast<float>(bouncy);
```
jocke1995, line 97: Should this scale with level of the upgrade? Or is the upgrade only supposed to be bought once?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeBounce.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "Components/RangeComponent.h"
+
+UpgradeRangeBounce::UpgradeRangeBounce(Entity* parentEntity) : Upgrade(parentEntity)
+{
+	SetName("UpgradeRangeBounce");
+	SetType(F_UpgradeType::PLAYER);
+
+	m_Price = 1000;
+	m_StartingPrice = m_Price;
+	m_MaxLevel = 1;
```
jocke1995, line 13: A bit expensive? :( Also same as above.. is this upgarde not supposed to scale with levels? Maybe I missed it, but I cannot find anything about level-scaling in this upgrade?

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeBounce.h
jocke1995: copy-paste comment

jocke1995 (changes requested):
Fix

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeVelocity.cpp
```diff
@@ -41,5 +41,5 @@ void UpgradeRangeVelocity::ApplyBoughtUpgrade()
 
 std::string UpgradeRangeVelocity::GetDescription(unsigned int level)
 {
-	return "Range Velocity: Increases velocity of the range attacks with 10 per level. Max velocity is 100, at upgrade level 5";
+	return "Range Velocity: Increases velocity of the range attacks with 30 per level. Max velocity is 300, at upgrade level 5";
```
kattalainen, line 44: Hold up! The overall velocity increase makes projectiles to sometimes go through walls. I was told because of the collision detection update rate(0.01?) max velocity should be 100 before collision detection starts to get iffy. Something has happened to the range velocity in the recent merges though so I have no idea what the limits should be now :> at 150 it SOMETIMES goes into walls, at 300 it almost always does. Not sure how to resolve this issue though?

kattalainen: I might add that I noticed that collision seems to work fine when the bouncy upgrade is bought, as in the projectiles no longer goes through walls even on 300

StortSpelProjekt/Game/src/Gamefiles/Components/RangeComponent.cpp
LeoWikstrom: It's only supposed to be bought once. Hence the max level 1. Could possibly make a longer time to live for higher levels. Would be easy to implement in the future if we feel it is needed

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeBounce.cpp
LeoWikstrom: If it's only buy-able once I think it should be expensive 😁

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeBounce.h
LeoWikstrom: Fuck...

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeRangeVelocity.cpp
LeoWikstrom: Does it go INTO walls or does it go THROUGH walls and out on the other side? IT should disappear when colliding with a wall, but due to a mismatch of bullet and our boundingboxes it doesn't always. Can look at a quick fix right away if necessary

kattalainen: Hmm ok then the "problem" is that the projectiles bounce on the walls without the upgrade? Idk lets take a couple of minutes and talk about this after lunch :>

kattalainen (approved):
when ranged velocity is 300 they SOMETIMES go through the walls. But idk, it's at an acceptable rate for me I guess. It could always be lowered if it causes a riot.

jocke1995 (approved):
nice

PR #225, opened 2020-11-12
More particle attributes

fjz345:


StortSpelProjekt/Engine/src/Particles/ParticleEffect.h
```diff
@@ -18,7 +19,7 @@ class Mesh;
 
 struct RandomParameter
 {
-	float2 interval = { 0.0f, 1.0f };
+	float2 interval = { 0.0f, 1.0f }; // default should be 0,0 rand does not work with 0,0
 	//rand distribution
 };
```
jocke1995: No need to have a struct with only 1 variable in it. But since you said that it will hold more data in the future, it is ok

Whixx (changes requested):
The particles flicker a LOT for both me and Björn when we're moving our character. This needs to be changed for the review to be approved. Also, i think the particles should not be connected to the players position other than the spawn location. When the player is moving it's clearly visible that the particles are following the player.

PR #230, opened 2020-11-17
Particle Play(), Stop() Not working as expected.

fjz345:
Test i used.

static int counterrr = 0;
        if (window->WasSpacePressed())
        {
            
            if (counterrr % 2 == 0)
            {
                partic->Stop();
            }
            else
            {
                partic->Play();
            }
            counterrr++;
        }


AntonAsbrink (approved):

jakkeez (approved):
Code looks good and works as intended

PR #231, opened 2020-11-17
Knockback upgrade

kattalainen:
Increases knockback with 5 per level. Also lowered base knockback to start at 5. Price starts at 100 and is increased with 100 per level.
Tip to get only this upgrade in the shop for testing: in UpgradeManager.cpp function fillUpgradeMap (line 137), comment out all other upgrades (line 141 to 189).


LeoWikstrom (dismissed):
Would have been nice with a specific icon in the shop. The default one is boring and should not be used. I would also have preferred a slightly more dramatic upgrade to the knockback. Five seems a biiit low 😅 Seems to work anyway

OtakuFreak13 (approved):
Code looks good and upgrade seems to work as intended.

LeoWikstrom (approved):
Cheeky icon change 😘

PR #232, opened 2020-11-17
Reroll Texture in Shop

jakkeez:
A reroll texture in shop has been added, it as no functionality. Price is subject to change


LeoWikstrom (changes requested):
I assume this task was just for the texture, since the actual functionality doesn't seem to work? Anyway, I couldn't run your code as it was. On line 31 in HealthComponent the name of the method is wrong. It should be ChangeMultiplicativeDamageReduction instead of ChangeMultiplicativeDamageReductionDamageReduction. On line 31 in UpgradeBlueJewel there's another mistake as it should again be ChangeMultiplicativeDamageReduction instead of ChangeProcentileDamageReduction. I guess this is a problem from the merge with develop

jakkeez:
As the base for this request is now in develop, the pull request has been reopened.

StortSpelProjekt/Game/src/Gamefiles/Shop.cpp
```diff
@@ -34,13 +34,13 @@ Shop::Shop()
...
 
-	// TODO: do this in InitScene BJ�RN
+	// TODO: do this in InitScene BJ�RN
 	EventBus::GetInstance().Subscribe(this, &Shop::OnShopGUIStateChange);
 }
 
 Shop::~Shop()
 {
-	// TODO: do this in UnInitScene BJ�RN
+	// TODO: do this in UnInitScene BJ�RN
```
Schtekt: It is a minor comment... but the "ö" character should be a part of it...

Schtekt (changes requested):
The code and the texture looks good in general.
I would however like to see the white spots/lines be removed from the "unmarked" texture before aproving this.

LeoWikstrom (approved):

Schtekt (approved):

PR #234, opened 2020-11-18
AI Back and Shoot

LeoWikstrom:
The demons now back away (while still shooting) when the player gets too close. Test if the range at which they start backing away is good, and if the speed with which they back seems reasonable


AntonAsbrink (dismissed):
Demons should not be moving and shooting at the same time as it will be weird with animations and i personally dislike it as balance. They should stand still for like 0.5 seconds while firing and then move before next shot. If it is implemented in this way i do belive they should move away a bit faster then.

jocke1995 (changes requested):
We discussed a bug. The range enemy keeps getting stuck walking forward and backwards non-stop. Otherwise I agree with Anton that the enemy should stop when shooting, to avoid problems with blended animations 👍

jocke1995 (approved):
Good code, Good AI, Good job

AntonAsbrink (approved):

PR #235, opened 2020-11-18
DualShock support

LeoWikstrom:
Support has been added for the controller DualShock 4.
Left stick or directional buttons to move.
Right stick to rotate camera.
X to jump.
Left trigger to sprint.
Right trigger to dash.
Left bumper for melee attack.
Right bumper for range attack.
Square to open shop


fjz345 (dismissed):
You will need to merge the newer develop and make sure everything still works and does not break anything.

Schtekt (dismissed):
Works great! would Actually ask to raise the sensitivity of both the mouse and the camera rotation though. it takes forever to turn 360 degrees ingame.

Schtekt (approved):

OtakuFreak13 (approved):
The game is much better with controller, so now I can't go back to palying with mouse and keyboard. Great Job!

PR #237, opened 2020-11-18
Selecting the shop by pressing F

jocke1995:
Look at the shop to "pick it", it should outline.
Then you can Press "F" to open/close the shop. the mouse should appear/dissapear as expected, and you should not be able to move/rotate the camera while in the shop-state.


jakkeez (approved):
Looks fine and works without any faults

LeoWikstrom (approved):
Lookin' goood 👌🏽

PR #239, opened 2020-11-18
Range damage Upgrade

AntonAsbrink:
It increases the range damage :/


StortSpelProjekt/Game/src/Gamefiles/UpgradeManager.cpp
```diff
@@ -201,6 +202,13 @@ void UpgradeManager::fillUpgradeMap()
 	upgrade->SetID(UPGRADE_KNOCKBACK);
 	// add the upgrade to the list of all upgrades
 	m_AllAvailableUpgrades[upgrade->GetName()] = upgrade;
+
+	// Adding KnockBack Upgrade
```
kattalainen: copy pasta

AntonAsbrink: I would never

kattalainen:
Don't know about the logic behind having an orange as the icon for the upgrade but besides that it seems to be working fine as intended :>

kattalainen (approved):

LeoWikstrom (approved):
I like the orange 👍🏽

PR #240, opened 2020-11-18
Melee attack speed Upgrade

kattalainen:
Increases melee attack speed with 20% (of base attack speed) per level. Starts out kinda cheap but gets expensive quickly because price doubles per level.
Tip to get only this upgrade in the shop for testing: in UpgradeManager.cpp function fillUpgradeMap (line 140), comment out all other upgrades (line 144 to 213).


StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeMeleeAttackSpeed.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+void UpgradeMeleeAttackSpeed::ApplyBoughtUpgrade()
+{
+	ApplyStat();
+}
+
+void UpgradeMeleeAttackSpeed::IncreaseLevel()
+{
+	m_Level++;
+	m_Price *= 2;
```
LeoWikstrom: Price must be based on level. Do not use *= or +=, only =

StortSpelProjekt/Game/src/Gamefiles/UpgradeManager.h
```diff
@@ -24,7 +24,8 @@ enum E_UpgradeIDs
 	UPGRADE_RANGE_BOUNCE = 6,
 	UPGRADE_POISON_ATTACK = 7,
 	UPGRADE_KNOCKBACK = 8,
-	UPGRADE_HEALTH_REGEN = 9
+	UPGRADE_HEALTH_REGEN = 9,
+	UPGRADE_MELEE_ATTACKSPEED = 10
```
LeoWikstrom: Should be 11 now

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -902,10 +902,10 @@ void ShopUpdateScene(SceneManager* sm, double dt)
...
     rotValue += 0.005f;
 
-    // Kod-p�kod-p�kod-p�kod-p�kod-l�sning
-    // Detta ska egentligen st� i "OnShopGUIStateChange" i Shop, men eftersom att vi inte har samma
-    // spelare i alla scener s� kan vi ej n� den aktiva spelaren i den scenen d�rifr�n.
-    // TODO: Flytta in den i den funktionen n�r vi har samma spelare i alla scener via Player::GetInstance().
+    // Kod-p�kod-p�kod-p�kod-p�kod-l�sning
+    // Detta ska egentligen st� i "OnShopGUIStateChange" i Shop, men eftersom att vi inte har samma
+    // spelare i alla scener s� kan vi ej n� den aktiva spelaren i den scenen d�rifr�n.
+    // TODO: Flytta in den i den funktionen n�r vi har samma spelare i alla scener via Player::GetInstance().
```
LeoWikstrom: Fiiix formatting

LeoWikstrom (changes requested):

StortSpelProjekt/Game/src/Gamefiles/Components/UpgradeComponents/Upgrades/UpgradeMeleeAttackSpeed.cpp
kattalainen: why though?

LeoWikstrom: Because otherwise the price won't reset when the player dies

kattalainen: not sure that's how it works but I'll change it anyway

LeoWikstrom: That is exactly how it works

kattalainen: Well I've tested it now and as I suspected none of the prices for any upgrades seem to be reset. Idk who was responsible for "fixing" that and should be burned at the stake but I might have an idea for a fix that I can patch in with next upgrade.

LeoWikstrom (approved):

AntonAsbrink (approved):

PR #241, opened 2020-11-19
Re-roll function in shop

jakkeez:
The re-roll is now implemented in the shop, the button costs 50 currency to use and is a flat price (i.e no price increase, this is subject to change) 
Try by beating a level and re-roll in the shop, also try to buy some upgrades and make sure that they do not reset down to zero if re-rolled.


LeoWikstrom (dismissed):

StortSpelProjekt/Game/src/Gamefiles/Shop.h
```diff
@@ -4,6 +4,8 @@
 #include "../Misc/EngineRand.h"
 #include "../Events/Events.h"
 
+#define REROLL_COST 50
+
```
fjz345: make it a variable

StortSpelProjekt/Game/src/Gamefiles/Shop.h
```diff
@@ -54,8 +56,9 @@ class Shop
 	void OnShopGUIStateChange(shopGUIStateChange* collisionEvent);
 
 private:
-	void upgradePressed(ButtonPressed* evnt);
+	void shopButtonPressed(ButtonPressed* evnt);
 	void sceneChange(SceneChange* evnt);
+	void updateShop();
```
fjz345: Other name than updateShop(), sounds like it should refresh the ui/logic of the shop but it also randomizes the inventory of the shop.

fjz345 (dismissed):
You mentioned that it is fixed 50, There is no task to change it later. I want you to implement a way to change the price with each re-roll. The exact price and price change can then be changed easily later. (ex, function called "rerollCost = rerollCostIncreaseFunction(int changedTimes)")

LeoWikstrom (dismissed):
Code looks good. Consider removing empty space on icon, as it is now possible to click outside the dark area of the button. Also, the price of a re-roll should be reset when the player dies, this should be done in Player's onResetGame function

fjz345 (approved):
Looking good, either you or the next person who changes the base price will have to change it in more than 1 place though, either const int ___ = 50 or #define ____ 50

LeoWikstrom (approved):

PR #242, opened 2020-11-19
It's a spider mah dudes

LeoWikstrom:
Added definition of spider to GameScene. It is not currently created in the factory, try spawning one manually to test it by adding  enemyFactory.SpawnEnemy("enemySpider") somewhere after enemyFactory.DefineEnemy("enemySpider", &spider)


StortSpelProjekt/Game/src/main.cpp
```diff
@@ -902,10 +923,10 @@ void ShopUpdateScene(SceneManager* sm, double dt)
...
     rotValue += 0.005f;
 
-    // Kod-p�kod-p�kod-p�kod-p�kod-l�sning
-    // Detta ska egentligen st� i "OnShopGUIStateChange" i Shop, men eftersom att vi inte har samma
-    // spelare i alla scener s� kan vi ej n� den aktiva spelaren i den scenen d�rifr�n.
-    // TODO: Flytta in den i den funktionen n�r vi har samma spelare i alla scener via Player::GetInstance().
+    // Kod-p�kod-p�kod-p�kod-p�kod-l�sning
+    // Detta ska egentligen st� i "OnShopGUIStateChange" i Shop, men eftersom att vi inte har samma
+    // spelare i alla scener s� kan vi ej n� den aktiva spelaren i den scenen d�rifr�n.
+    // TODO: Flytta in den i den funktionen n�r vi har samma spelare i alla scener via Player::GetInstance().
```
fjz345: Bra ändring

kattalainen (commented):
Looking good except for a crash that I suspect might be resolved when spider is handled with enemyfactory. It crashes when I kill all enemies -> teleport to shop -> teleport back to game -> crash. But it only happens when the spider is spawned, AND killed.

fjz345 (approved):

StortSpelProjekt/Game/src/main.cpp
kattalainen: haha den skiten kommer hänga med ett tag nu

kattalainen (approved):

PR #244, opened 2020-11-19
ID:140 Camera Jump Fix

fjz345:
Just fixed so it doesn't jump, the underlying problem is still unknown


kattalainen:
problem is still there for me :< I guess it's a slight improvement though, hard to tell really

LeoWikstrom (changes requested):
Does not work

PR #246, opened 2020-11-20
Spider AI

LeoWikstrom:
Added an AI flag, F_AI_FLAGS::RUSH_PLAYER which, if set, makes the enemy rush the player rather than circle her. To test, set this flag in the definition of one of the enemies, e.g. the zombie (as the spider is not yet created by the factory)


jocke1995 (approved):
Tried all flags together. Works as expected.

fjz345 (approved):

PR #247, opened 2020-11-20
Fixed shop reset

LeoWikstrom:
There was a function call missing which would reset the level of the upgrades in the shop. This call was added, and included a reset of the price.
There was also a problem with the order of "apply upgrade" and "increase level" for the upgrades. This was fixed


jocke1995 (approved):
Works, cant find any issues.

fjz345 (approved):

PR #251, opened 2020-11-23
Particle Emitter can have multiple effects

fjz345:


StortSpelProjekt/Game/src/Gamefiles/Components/PoisonDebuff.cpp
```diff
@@ -26,7 +26,9 @@ component::PoisonDebuff::PoisonDebuff(Entity* parent, int damagePerTick, int tic
 	settings.randRotationSpeed = { 0, 3 };
 
 	Texture2DGUI* particleTexture = static_cast<Texture2DGUI*>(AssetLoader::Get()->LoadTexture2D(L"../Vendor/Resources/Textures/Particles/poison_particle.png"));
-	parent->AddComponent<component::ParticleEmitterComponent>(particleTexture, &settings, true);
+	settings.texture = particleTexture;
+
+	parent->AddComponent<component::ParticleEmitterComponent>(&settings, true);
```
jocke1995, line 31: Separate constructor for this is nice, but what happens if I want to add another particle effect to this component somewhere else? I cannot find any function calls to append a particleEffect to an existing component. Is the only way to add an effect to create all the effects directly in the constructor?

jocke1995 (changes requested):
Some questions about appending particle effects to a component without calling constructor.

jocke1995 (approved):

AntonAsbrink (approved):

PR #252, opened 2020-11-23
Portal Particle effect

fjz345:
As decided in the morning meeting, simple effect, added a color variation (yellow) which can be changed by changing texture path from _blue to _yellow. I thought blue was better.


jocke1995 (dismissed):
Looks ok, I can sense problems incoming when we're moving the teleport around, when its no longer at the same positon everytime. But I guess this will work for now, and later when fredriks maps are in game, we simply change these positions to be around the teleport.

LeoWikstrom (dismissed):
Good job. If it is now possible to have more than one particle effect on an  entity, how about both blue AND yellow?

LeoWikstrom (approved):

jocke1995 (approved):

PR #259, opened 2020-11-25
Melee attack range upgrade

kattalainen:
Scales the melee hitbox. Max 3 levels. 
Picture with the hitbox at different levels: https://drive.google.com/file/d/1TY2k6Wskikz55x26oPss_trY-5nPODB4/view?usp=sharing. Comment out lines 147 to 236 in UpgradeManager.cpp to only get this in the shop. Bring cash cause it's kinda expensive.

Also snuck in a little fix for "death duration" timer to reset on death, so you won't instantly die on hp 0 after restarts.


jocke1995 (changes requested):
The particles of the teleport seems to be restarting or somethign once every second. I dont know if its something you did or if its on develop, but we will talk about it.

kattalainen:
encrypted models are now required to be downloaded for the game to run.

jocke1995 (approved):
works

Whixx (approved):
Seems to work fine

PR #261, opened 2020-11-25
Reworked camera rotation

fjz345:
Now uses quaternions and also cleaned up the code. Also added mouse sensitivity in options. Hopefully there is no more "jumping" when testers test.


kattalainen (approved):
It's working now! smooth as a dolphin!

Whixx (approved):
Smoooooooth

PR #262, opened 2020-11-25
Smooth Animation Transitions

Whixx:
Should be fully testable from game. The following animations should be visible:
Idle (when idle duh), Running (on movement), Axe Swing (on melee attack), Left hook (on ranged attack).

The animated model bug where some parts of the model were translated to origin is also solved.


StortSpelProjekt/Engine/src/Renderer/AnimatedModel.h
```diff
@@ -31,19 +33,36 @@ class AnimatedModel : public Model
...
 
-    double m_Time;
-    bool m_AnimationIsPaused = true;
-    Animation* m_pActiveAnimation;
+    bool m_AnimationIsPaused = false;
+    double m_BlendTransitionTime = ANIMATION_TRANSITION_TIME;
+    double m_BlendTimeElapsed = 0;
+    std::map<std::string, TransformKey> m_BlendAnimationState;
+    Animation* m_ReactivateAnimation;
+    Animation* m_QueuedAnimation;
```
OtakuFreak13: m_p Should be used for pointers.

OtakuFreak13 (changes requested):
Just fix the minor code standard issue and I will approve.  Nicely done Wille!

kattalainen (commented):
Seems fine, besides range animation only playing once when right mouse button is held down (as we discussed), the only issue I have is with the melee hitbox shrinking. Lets try and find a way to fix that before merge into develop.

kattalainen (dismissed):

kattalainen (approved):

LeoWikstrom (approved):
Some weird stuff with the animations, but the transitions are smooooth. Good job

OtakuFreak13 (approved):

PR #275, opened 2020-11-30
Removed turn rate

LeoWikstrom:
Made player not helicopter by removing turn rate

### DIFF
--- a/StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
+++ b/StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
@@ -48,9 +48,6 @@ component::PlayerInputComponent::PlayerInputComponent(Entity* parent, unsigned i
 	m_UpdateShootId = -1;
 	m_UpdateDashId = -1;
 
-	m_DegreesToTurnTo = 0;
-	m_RadiansToTurn = 0;
-
 	m_TurningTimer = 0.0f;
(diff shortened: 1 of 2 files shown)


StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
```diff
@@ -48,9 +48,6 @@ component::PlayerInputComponent::PlayerInputComponent(Entity* parent, unsigned i
 	m_UpdateShootId = -1;
 	m_UpdateDashId = -1;
 
-	m_DegreesToTurnTo = 0;
-	m_RadiansToTurn = 0;
-
 	m_TurningTimer = 0.0f;
 	m_TurningInterval = 0.0f;
```
fjz345, line 52: Looks like this should also get removed? Can't find any logic with these variables, they are only set to stuff.

fjz345 (dismissed):
Thought the task was to fix the turn rate, not remove it.

StortSpelProjekt/Game/src/Gamefiles/Components/PlayerInputComponent.cpp
LeoWikstrom: There is a logic to them, you just don't understand it. They have less than nothing to do with this task

fjz345 (approved):

jocke1995 (approved):

PR #276, opened 2020-11-30
Particle bunched up fix

fjz345:
1. Particles not bunched up
2. Dt is close to 0 on change scene
3. Fredriks range particle is changed

### DIFF
--- a/StortSpelProjekt/Engine/src/ECS/SceneManager.cpp
+++ b/StortSpelProjekt/Engine/src/ECS/SceneManager.cpp
@@ -105,8 +105,10 @@ Scene* SceneManager::GetScene(std::string sceneName) const
     return nullptr;
 }
 
-void SceneManager::ChangeScene()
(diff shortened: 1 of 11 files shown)


jocke1995 (approved):
Perfect! Good job my man!

StortSpelProjekt/Engine/src/ECS/SceneManager.cpp
```diff
@@ -105,8 +105,10 @@ Scene* SceneManager::GetScene(std::string sceneName) const
     return nullptr;
 }
 
-void SceneManager::ChangeScene()
+bool SceneManager::ChangeScene()
 {
+	bool changedScene = false;
```
Schtekt, line 110: Why not set changedScene to the value of m_ChangeSceneNextFrame?

Schtekt (approved):
Looks good!
Have a minor comment which is not really necessary to attend to :)

PR #283, opened 2020-12-01
Zombie sounds NOT DEMON. Download sounds from drive!

Schtekt:
Zombies now grunt and use a new "on hit" sound effect.
Make sure to download the sound files from the drive and place them in Vendor/Resources/Audio/IgnoredAudio.


LeoWikstrom (changes requested):
looks and sounds good, but crashes when enemyDemon (and presumably enemySpider, though unconfirmed) tries to grunt

LeoWikstrom (approved):

jakkeez (approved):

PR #284, opened 2020-12-01
Added sound when interacting with shop

LeoWikstrom:
Made the shop keeper say "Hello there" when the player interacts with the shop. Also added Obi-Wan Kenobi as shop keeper. Can remove if the reviewer so wishes. DON'T FORGET TO DOWNLOAD THE LATEST ANIMATED AND ENCRYPTED MODELS FROM THE DRIVE


fredrix96 (changes requested):
The sound needs to be encrypted :)

AntonAsbrink (changes requested):
Is this not copyrighted?

AntonAsbrink (changes requested):
Still copyrighted material, don't use obi-wan or a sound clip from a movie

PR #286, opened 2020-12-01
Remove enemies which are outside the map

fredrix96:
Paste this:
`t->SetPosition(comps->pos.x, -5.0f /*comps->pos.y*/, comps->pos.z);`
on row 206 in EnemyFactory.cpp.
If they all die, then it works

### DIFF
--- a/StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.cpp
+++ b/StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.cpp
@@ -23,6 +23,7 @@ EnemyFactory::EnemyFactory()
 	m_SpawnCooldown = 1;
 	m_MinimumDistanceToPlayer = 10;
 	m_SpawnTimer = 0.0f;
+	m_TotalTime = 0.0f;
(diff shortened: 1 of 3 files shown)


StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.cpp
```diff
@@ -418,7 +420,30 @@ void EnemyFactory::Update(double dt)
 			m_EnemiesToSpawn--;
 			m_EnemySlotsLeft--;
 		}
+
+		// After a certain time, check if enemies are outside the map
+		// If they are, remove them
+		if (static_cast<int>(m_TotalTime) % 10 == 0)
+		{
+			m_TotalTime += 1.0f;
```
jocke1995: Use a float insttead to avoid calculations with modulo. In that case, you dont have to add +1.0 aswell. Otherwise, great work :)

jocke1995 (changes requested):
Pls fix some minor issues :) otherwise, great work!

jocke1995 (approved):

kattalainen (approved):

PR #290, opened 2020-12-02
Fix for fps and color change

Schtekt:


fredrix96 (changes requested):
I am sad that the braziers are GONE, however, the particles are wrong for all chandeliers except the one in the SpawnRoom. Besides that, I almost gained 100fps, so that is nice, and I also like the colors of the lights

LeoWikstrom (dismissed):
Agree with Fredrik. Looks great, good FPS, but make sure all Chandeliers have the same particle effect. Either completely gone or completely there, not in between

kattalainen (approved):

fredrix96 (approved):

PR #291, opened 2020-12-02
Press f in shop.

fjz345:
Stone slab that says "Press 'f' to Interact" in the shop.


LeoWikstrom (dismissed):
Firstly, since you have elected to use a small letter 'f', the sign becomes very unclear and hard to read. A capital 'F' would be more readable and better fit the purpose of the sign.
Secondly, the sounds in IgnoredAudio should NOT be uploaded to git. Might be okay since they are encrypted, but not good practice.

jakkeez (dismissed):
Change scale on the model to 1.4 so the player read the text

jakkeez (approved):

LeoWikstrom (approved):
Better

PR #293, opened 2020-12-02
Healthbars on enemies

LeoWikstrom:
Healthbars appear when an enemy spawns and when it takes damage. They disappear after three seconds. Also fixed so enemies are not killed by one attack


jakkeez (approved):

fredrix96 (approved):
I love the healthbars, good job!

PR #301, opened 2020-12-02
Upgrade Balance and fix

AntonAsbrink:
Blue Jewel: This upgrade is insanely expensive for what it does. It gives 12% more effective health which translates to 60hp without boosters. For 900 curr it is too expensive.
- Damage reduction: 40% -> 25%
- Health Threshold: 30% -> 50%
- Cost: 900 -> 200
It now has 12.5% effective health boost for a lot less cost.

Health Regeneration: It gives way to much health. A round takes about 50-70 seconds early on which translate at 4hp/s to about 200-280 more hp.
- Health Regen: 4 -> 2
It now has a effective health off 100-140 which is more inline with health boost.

Poison Attack: This upgrade had a flat damage over time which is hard to interpert when the player dont know the enemies health or their own damage. Instead it does a procentile amount of range damage over time.
- Damage per tick: 6 -> 4% of range damage
- Damage Scaling: 2 -> 4% of range damage
- Fixed Description
This gives 40% of range damage on first level, 88% on second, 144% on third etc..

Melee Radius: This QoL upgrade is very nice for game feel and shouldn't take the entire first round currency to purchase.
- Cost: 400 -> 250

### DIFF
--- a/StortSpelProjekt/Game/src/Gamefiles/Components/HealthComponent.cpp
+++ b/StortSpelProjekt/Game/src/Gamefiles/Components/HealthComponent.cpp
@@ -139,6 +139,10 @@ void component::HealthComponent::TakeDamage(int damage)
 		m_pParent->GetComponent<component::UpgradeComponent>()->OnDamage();
 	}
 	damage = (damage - m_FlatDamageReduction) * m_MultiplicativeDamageReduction; //Flat Damage gets applied first followed by multaplicative damage
+	if (damage <= 0)
+	{
+		damage = 1;
+	}
 	ChangeHealth(-damage);
 	//Damage reduction stat is reset to allow upgrade to change again
 	m_FlatDamageReduction = 0.0;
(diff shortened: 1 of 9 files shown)


kattalainen (dismissed):

kattalainen:
I like the changes and will approve. But if you have the time could you please change the icons for rangedDamage, rangeLifeSteal and rangeVelocity to use the RangeUpgrade.png instead of default or the orange :> 
Edit: nvm did it myself

kattalainen (dismissed):

OtakuFreak13 (dismissed):
The Blue Jewl has a fault in its description as it says that it reduces damage taken by x% when it actually makes the damage taken x% of original. So 100 damage with 25% of this upgrades reduction means you take only 25 damage not 75 damage as the description makes it out to be. 

Also since blue jewel makes it so you only take 6% damage on the second level and then 1% and finally 0% on the 4th level, we are imortal for a pretty cheap price. 

Maybe add a maxLevel to the upgrade so we can only buy it once? Or maybe change how it works so it works like the description says it does.

OtakuFreak13 (dismissed):
When increasing the level you actually get less damage reduction since it has "m_DamageReduction = pow(..." and not "m_DamageReduction += pow(..." so you don't get an increase in damage reduction but a decrease.

OtakuFreak13 (approved):

kattalainen (approved):

PR #303, opened 2020-12-02
Dash Sound

kattalainen:
Added sound for dashing. Also changed values for player in shopscene to make it more consistent.


OtakuFreak13 (dismissed):

fredrix96 (approved):
Seems to work good. I like the new dash sound and the coin is beautiful!

OtakuFreak13 (approved):

PR #309, opened 2020-12-04
Pillars, beams and banners

LeoWikstrom:
Added pillars between walls, ceiling beams between rooms, and fixed z-fighting for the banners


AntonAsbrink (changes requested):
One beam is Z-fighting
Roof is clipping with beam on outside room

Otherwise looks nice

Schtekt (changes requested):
Generally looks good. I have a few problems though.
The beams that are in the ceiling seem to stretch their textures. Might just be my eyes tricking me but please take a look at that.

The beam between an empty room and a ramp room (in this case) seem to have light on it, or letting the ceiling through. See the attached image.
image

Schtekt (approved):

AntonAsbrink (approved):

PR #311, opened 2020-12-04
GameOver stats

AntonAsbrink:
Game over shows:
-Level Reached
-Enemies Killed
-Time Played

Added console command
-"KillPlayer" Sets the players health to 0 and resulting in death
-"SkipLevel" Publish LevelDone event

Made GameGUI to a singleton to more easily access it in gameoverhandler. It was previously being used as a global variable in main only and suits well as a singleton.


Schtekt (dismissed):
Looks good!

jocke1995 (dismissed):
Looks good, just need to move the text a bit to the right

Schtekt (approved):

jocke1995 (approved):
niceöh

PR #312, opened 2020-12-04
New shop

Schtekt:
The shop room/environment has been updated :)


jocke1995 (dismissed):
Requests:
1. More lights on each wall
2. Change the direction of the moonLight so it fits the moon.
3. Also make sure that the shadow-map from the moon is locked and adjusted for the shop scene.

kattalainen (commented):
I agree with Jocke about the lighting. 
Another thing is that when we spawn in the shop we are very close to the teleporter, and it might be too easy to accidentally move/dash into it.

kattalainen (dismissed):
Maybe still a little bit too dark around the shop itself, a light source above the shop might make it better but it's good enough for me :>

kattalainen:
not sure I like the candles, but that was part of another task right?

edit: maybe changing the color values of the light to match the rest of the environment might make it more fitting? it kind of sticks out imo.

kattalainen:
Nevermind the lights, I think what I had a problem with is the tomb in the shop scene. Maybe put a "real" light source closer to it so it doesn't look like a dark void. The fact that the candles don't emit any light makes it look weird when they are in a very dark place such as the corner in the shop scene :> 
But you probably have better things to do atm, tell me your thought's and I'll probably just approve.

LeoWikstrom (changes requested):
I personally think it looks weird without a shopkeeper. Might also be a good idea with a light on the "press f" sign? Would also like Stefan and Hasse banners 😁

LeoWikstrom (changes requested):
Love the goblin. Would like it if you either added the goblin to the shop model (so both of them get highlighted) or move the highlight to the goblin. Otherwise, great job 👍🏽

kattalainen:
I think you went a bit overkill with the stefan and hans banners, check discord for rest of groups opinions

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -953,132 +935,39 @@ Scene* ShopScene(SceneManager* sm)
...
+    //entity = scene->AddEntity("pressf");
+    //mc = entity->AddComponent<component::ModelComponent>();
+    //mc->SetModel(pressfModel);
+    //mc->SetDrawFlag(FLAG_DRAW::DRAW_OPAQUE | FLAG_DRAW::GIVE_SHADOW);
+    //
+    //tc = entity->AddComponent<component::TransformComponent>();
+    //tc->GetTransform()->SetPosition(-25.0f, 0.0f, 18.0f);
+    //tc->GetTransform()->SetRotationY(PI - PI / 4);
+    //tc->GetTransform()->SetScale(1.4);
+    //tc->SetTransformOriginalState();
```
kattalainen: I think you forgot to remove this.

kattalainen (commented):
I gotta say I don't really think the meme banners are fitting but I guess this is a good compromise and others are fine with it.
Just remove commented code and I'll approve

LeoWikstrom (dismissed):
I wanted Stefan more to the left 😞

StortSpelProjekt/Game/src/main.cpp
Schtekt: Done

LeoWikstrom (dismissed):

LeoWikstrom (approved):

jakkeez (approved):
Works as intended. I dont mind the Stefan banner to be honest. But I would like to bring this up infront of everyone if this is such a split opinion. 
I will approve

fredrix96 (approved):
Cozy shop

PR #314, opened 2020-12-04
Added Light Intensity

jocke1995:
Colors should now be between 0-1, and if we want a higher intensity on the lightcolors, we simply change the Intensity. This will make it easier to adjust the light-flickering which the torches should have.

### DIFF
--- a/StortSpelProjekt/Engine/src/ECS/Components/Lights/PointLightComponent.cpp
+++ b/StortSpelProjekt/Engine/src/ECS/Components/Lights/PointLightComponent.cpp
@@ -69,8 +69,9 @@ namespace component
 		return m_pPointLight;
 	}
 
-	void PointLightComponent::UpdateLightColor()
+	void PointLightComponent::UpdateLightColorIntensity()
 	{
 		m_pPointLight->baseLight.color = m_pBaseLight->color;
+		m_pPointLight->baseLight.intensity = m_pBaseLight->intensity;
 	}
 }
--- a/StortSpelProjekt/Engine/src/Renderer/HLSL/LightCalculations.hlsl
+++ b/StortSpelProjekt/Engine/src/Renderer/HLSL/LightCalculations.hlsl
@@ -93,7 +93,7 @@ float3 CalcDirLight(
 		shadow = CalculateShadow(fragPosLightSpace, dirLight.textureShadowMap);
 	}
 
-	DirLightContribution = (kD * albedo / PI + specular) * radiance * NdotL;
+	DirLightContribution = (kD * albedo / PI + specular) * radiance * NdotL * dirLight.baseLight.intensity;
 	return DirLightContribution * (1.0f - shadow);
 }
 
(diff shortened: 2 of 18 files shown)


StortSpelProjekt/Engine/src/Renderer/HLSL/LightCalculations.hlsl
```diff
@@ -93,7 +93,7 @@ float3 CalcDirLight(
 		shadow = CalculateShadow(fragPosLightSpace, dirLight.textureShadowMap);
 	}
 
-	DirLightContribution = (kD * albedo / PI + specular) * radiance * NdotL;
+	DirLightContribution = (kD * albedo / PI + specular) * radiance * NdotL * dirLight.baseLight.intensity;
```
fjz345, line 96: Would like radiance normalized since intensity now scales the color up. (2.0, 0, 1.0) should be == (1.0, 0, 0.5)

StortSpelProjekt/Engine/src/ECS/Components/Lights/PointLightComponent.cpp
```diff
@@ -69,8 +69,9 @@ namespace component
 		return m_pPointLight;
 	}
 
-	void PointLightComponent::UpdateLightColor()
+	void PointLightComponent::UpdateLightColorIntensity()
 	{
 		m_pPointLight->baseLight.color = m_pBaseLight->color;
+		m_pPointLight->baseLight.intensity = m_pBaseLight->intensity;
 	}
```
fjz345, line 76: Do not see the reason why this is not in Light class, it has the same code in all lights.

fjz345 (changes requested):
Do fix.

StortSpelProjekt/Engine/src/ECS/Components/Lights/PointLightComponent.cpp
fjz345: Also not sure why it is not called Update(), or just UpdateLight.

fjz345: Not sure what the point of this function is either, it is called on SetIntensity(), then you set the color which is unchanged in this function. Also don't quite understand the baseLight vs m_pBaseLight.

StortSpelProjekt/Engine/src/Renderer/HLSL/LightCalculations.hlsl
fjz345: (Does not have to be normalized in the shader)

StortSpelProjekt/Engine/src/ECS/Components/Lights/PointLightComponent.cpp
jocke1995: It is in the baselight

StortSpelProjekt/Engine/src/Renderer/HLSL/LightCalculations.hlsl
jocke1995: now normalised

fjz345 (approved):
GJ

AntonAsbrink (approved):

PR #315, opened 2020-12-04
Zombie attack sounds

kattalainen:
GET NEW WAV FILES FROM THE IGNOREDAUDIO.ZIP!
Changed the earlier zombie grunt sounds to be their attack sounds. Replaced the grunt sounds with new files.


StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.cpp
```diff
@@ -185,34 +188,39 @@ Entity* EnemyFactory::Add(const std::string& entityName, EnemyComps* comps)
...
+	{
+		ae->AddVoice(comps->onAttackSounds[0], L"OnAttack");
 	}
 
-	if (comps->attackSound.size() > 0);
+	if (comps->walkSound.size() > 0)
 	{
-		ae->AddVoice(comps->attackSound, L"Attack");
+		ae->AddVoice(comps->walkSound, L"Walk");
 	}
```
Schtekt: I would like to see that wals sounds follows the same formula as all the other sounds (Randomize if there is more than one option of sound).

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -175,16 +175,24 @@ Scene* GameScene(SceneManager* sm)
     Texture* currencyIcon = al->LoadTexture2D(L"../Vendor/Resources/Textures/2DGUI/currency.png");
 
     AudioBuffer* bruhVoice = al->LoadAudio(L"../Vendor/Resources/Audio/bruh.wav", L"Bruh");
-    AudioBuffer* playerHit1 = al->LoadAudio(L"../Vendor/Resources/Audio/Femalegrunt.wav", L"PlayerHit1");
+    //AudioBuffer* playerHit1 = al->LoadAudio(L"../Vendor/Resources/Audio/Femalegrunt.wav", L"PlayerHit1"); // this sound starts immediately, while the delayed starts after around 0.25 seconds
```
Schtekt: Is this intended for use? if not, remove it.

StortSpelProjekt/Game/src/main.cpp
```diff
@@ -296,14 +304,19 @@ Scene* GameScene(SceneManager* sm)
...
+    zombie.onGruntSounds.emplace_back(L"ZombieGnarl3");
+    zombie.onGruntSounds.emplace_back(L"ZombieGnarl4");
+    zombie.onGruntSounds.emplace_back(L"ZombieGnarl5");
+    zombie.onGruntSounds.emplace_back(L"ZombieGnarl6");
+    zombie.onGruntSounds.emplace_back(L"ZombieGnarl7");
+    zombie.onAttackSounds.emplace_back(L"ZombieAttack1");
+    zombie.onAttackSounds.emplace_back(L"ZombieAttack2");
+    zombie.onAttackSounds.emplace_back(L"ZombieAttack3");
+    zombie.onAttackSounds.emplace_back(L"ZombieAttack4");
+    zombie.onAttackSounds.emplace_back(L"ZombieAttack5");
```
Schtekt: Might be a mistake from me that has followed since, but the onHit emplacement should not be indented here, if it is intended to be separated from the rest use a newline instead.

Schtekt (changes requested):
A few things I would like to see be changed or adressed, but in general good.

fjz345 (dismissed):
I think you should lower the volume a little bit on the sound, hurts my ears a bit with my super headphones. (relative to the character jump sound and the zombie sound is a lot harsher overall.)

Schtekt (dismissed):

jakkeez (approved):

Schtekt (approved):

PR #316, opened 2020-12-07
Camera Stutter Fix

AntonAsbrink:
This feature fixes the camera stutter. This is done instead of removing renderupdate and keeping our current timestep solution. Because that is the proper and best way off handling timestep.

I also changed so renderPosition is linearly interpolated between update frames.


LeoWikstrom (dismissed):
Seems to work, can't find anything wrong with the code. Good job 👍🏽

LeoWikstrom (approved):

jakkeez (approved):

PR #318, opened 2020-12-07
Shiny gold on banners, vertices clean-up and no more clipping with the walls

fredrix96:
The banners are no longer clipping with the walls and look better in general, due to the fact that I had to remove colliding faces in blender. I have also added golden rings along the edges to make the banners shine a little whenever a light touches them. However, there is still one artifact left, but it has to do with the fact that lights are able to shine through walls and has nothing to do with the model.


Schtekt (dismissed):

kattalainen:
Skärmbild 2020-12-08 104945
There is a lot of this kind of light glitch flickering. Idk if it always existed but should probably fix this before merge.

jakkeez (approved):

kattalainen (approved):

PR #319, opened 2020-12-07
Spider sound effect

jakkeez:
A new spider sound effect, new ignored audio file up on drive. 
Good if zombie attack sound pull-request comes before this one, as it is based of it
Edit: There is also an update to the license file, as new sounds has been added that do require a mention. Check it as well


Schtekt (dismissed):
Would like to see some changes to the sound itself. When multiple spiders are attacking the player it sounds more like lava burns from minecraft?

LeoWikstrom (dismissed):

LeoWikstrom (approved):

Schtekt (approved):

PR #320, opened 2020-12-08
Stop Moving You're Dead

AntonAsbrink:
Unsubcribes from everything in playerinput to prevent the player from performing actions when they have died.

The player can still do animations but  apparently a very nice fix will automagically work with Williams animation implementation in his branch so lets call this good enough eyy?


StortSpelProjekt/Game/src/Gamefiles/Components/MeleeComponent.cpp
```diff
@@ -13,6 +13,7 @@
 #include "../ECS/SceneManager.h"
 #include "../Renderer/Transform.h"
 #include "../ECS//Components/TemporaryLifeComponent.h"
+#include "../Player.h"
```
Schtekt: This include is not used. Please remove it.

Schtekt (dismissed):

LeoWikstrom (dismissed):
Doesn't work properly. Animations are still played when dead. When player is killed by enemies (instead of by KILLPLAYER) she doesn't fall over

LeoWikstrom (approved):
Is fine

Schtekt (approved):

PR #321, opened 2020-12-08
Pause menu

LeoWikstrom:
Pause now has a menu with three options: Return to game, which returns you to the game, Main Menu, which sends you to the main menu, and Exit, which exits the game. Hard to comprehend, I know


fredrix96 (changes requested):
Closeby enemies can still hurt you in the pause menu and effects such as the poison keep lowering the enemies healths

fredrix96 (approved):
Seems to work. Good job

OtakuFreak13 (approved):
Works as intended. Only thing that might be improved is that buttons don't get highlighted in the pause menu.

PR #322, opened 2020-12-08
ASAP broken teleporter fix

AntonAsbrink:
A bug came in were sometimes the teleporter would not work in one way or another. Causing the player to constantly be sent back and forward until crash. Instantly sending the player back to shop or back to play map.

The problem was twofold. One oldposition was not defined and changed to position in SetPosition function. This caused it to be (0, 0, 0). Render position then sometimes interpolated between (0, 0, 0) and the set position.
Second problem WorldMatrix was using render position and was being used for handling collision. So sometimes the world position would sometimes be (0, 0, 0) when concering collision.

### DIFF
--- a/StortSpelProjekt/Engine/src/ECS/Components/BoundingBoxComponent.cpp
+++ b/StortSpelProjekt/Engine/src/ECS/Components/BoundingBoxComponent.cpp
@@ -69,7 +69,7 @@ namespace component
 				obb.Orientation = m_OriginalBoundingBox.Orientation;
 
 				// then do all the transformations on this temoporary OBB so we don't change the original state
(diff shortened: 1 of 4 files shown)


OtakuFreak13 (changes requested):
Fixed the teleport problem for me. Have tried getting to shop 3 times, each time shutting down the game so I got a fresh start, and they all worked.

But particle effects when enemies dies happen at the teleporters position which might be origo which I don't think they did before. 
Edit: We decided that this about particle will be done in another fix so I aprove.

jakkeez (approved):
Issue with the particles, however, this is seemingly not from this request.

OtakuFreak13 (approved):

PR #323, opened 2020-12-08
Audio listener fix

kattalainen:
Fixed an audio bug that Tim found last week. 
Changed the listener forward and up vectors to be fetched from the camera rather than the player. If you would spam A-D when sounds were playing the sound would jump from left to right. Using the camera vectors feels better.

Also snuck in a little initialisation of jump variable so player wont play the jump sound the first time she steps on the skeletons or the ramp.


OtakuFreak13 (approved):
Seems to work as intended

OtakuFreak13 (approved):

fredrix96 (approved):

PR #327, opened 2020-12-08
Smaller Textures

AntonAsbrink:
A lot of textures has been reduced to generally 512x512
Take a look at all textures and see if the look good.

This is the following models but might not be all
-Tree
-PalisadeWall(Outside wood wall)
-Tomb
-Zombie
-Skulls
-CubePBR(not in game)
-Vase
-Statue
-Pillar
-Anvil
-SwordPile
-Teleporter


LeoWikstrom (approved):
Doesn't change performance for me, but also doesn't look noticeably different (except the palisade, which looks better). Good job

jakkeez (approved):

PR #328, opened 2020-12-08
Moved text during time rounds

fredrix96:
Make sure that the text looks "better" on the enemy gui (to the upper left) during time rounds and at the end of a time round


LeoWikstrom (approved):
Small change, but looks right

AntonAsbrink (approved):

PR #330, opened 2020-12-09
Even more memory saved

AntonAsbrink:
Found some more areas with excessive sizes.
Made particles smaller to generally be 256x256 instead of 512x512. Aswell as removing some empty space around them.

Removed some unused models and textures. Reduced wooden floor texture size.


LeoWikstrom (approved):
Still no difference for my fps, but VRAM and RAM are both definitely down. Good job 👍🏽

fredrix96 (approved):
Damn, you have saved us a lot of memory now

PR #331, opened 2020-12-09
"Trying to remove entity ' ' that does not exist in scene" fix

LeoWikstrom:
Fix to keep program from trying to remove an entity that has already been removed. Was caused by "RemoveMe" being sent several times by an entity in one frame. Fixed by checking if entity is already listed to be removed


jocke1995 (dismissed):
Wow! much wow!

jocke1995 (approved):
Didn't quite understand the variable name.. its ok for this time, but next time please describe it a bit more.

fredrix96 (approved):

PR #334, opened 2020-12-09
Lower music volume and there is music option now

fredrix96:
Make sure that the music volume is okay and that you can mute the music via the option (Note, this option does not require a restart of the application)


OtakuFreak13 (approved):

jakkeez (approved):
Works and no issue with the code as I can see

PR #337, opened 2020-12-09
Fixed particles

AntonAsbrink:
Fixes particles sometimes spawning in origo(Transform did that one)
Added checks to prevent particles spawning on scenechange


OtakuFreak13 (approved):
Tested and particle on death effects works correctly now, as it spawns on the body and also does not spawn when you've teleported to the shop scene. Explosion and poison particles also works as intended.

LeoWikstrom (approved):

PR #338, opened 2020-12-09
Merchant now greets the player!

Schtekt:
This pull request is based of the code of the new shop PR, do not review code from that, only the sounds themselves and the method of playing them.

EDIT:
Also added the player walk sound to this PR as it was so little code :)


StortSpelProjekt/Game/src/main.cpp
```diff
@@ -700,6 +700,12 @@ Scene* ShopScene(SceneManager* sm)
     TextureCubeMap* skyboxCubemap = al->LoadTextureCubeMap(L"../Vendor/Resources/Textures/CubeMaps/skymap.dds");
 
 	AudioBuffer* music = al->LoadAudio(L"../Vendor/Resources/Audio/shopMusic.wav", L"ShopMusic");
+    AudioBuffer* gawblinSound = al->LoadAudio(L"../Vendor/Resources/Audio/IgnoredAudio/HelloThere.wav",L"HelloThere");
```
LeoWikstrom: Doesn't work, wrong path

LeoWikstrom (changes requested):

StortSpelProjekt/Game/src/main.cpp
LeoWikstrom: Also, sounds need to be louder. They sound good, but can barely be heard

LeoWikstrom (dismissed):
Love the voice 👍🏽

fredrix96 (changes requested):
I love the added sounds, but I still think that the volume is too low. However, the walk sound seems to sometimes suddenly stop playing after fighting or jumping. It also seems to be playing during jumps

LeoWikstrom (approved):

fredrix96 (approved):
Seems fine now!

PR #340, opened 2020-12-10
All entities can and are animated

Whixx:
Zombie: Attack, run.
Player: Ranged attack, melee attack, run, idle.
Spider: Attack (bite), run.
Demon: Walk, Idle, Attack (ranged).


StortSpelProjekt/Engine/src/ECS/Components/AnimationComponent.h
```diff
@@ -18,16 +25,51 @@ namespace component
...
+        friend class ShadowRenderTask;
+        
+        AnimatedModel* m_pAnimatedModel = nullptr;
+        SkeletonNode* m_pSkeleton = nullptr;
+
+        DirectX::XMFLOAT4X4 m_GlobalInverseTransform;
+        std::map<std::string, TransformKey> m_AnimationState;
+
+        std::vector<Animation*> m_Animations;
+        std::pair<Animation*, AnimationInfo> m_pPendingAnimation;
```
fredrix96: I am pretty sure you don't need a p in the names here

fredrix96 (changes requested):
The demon's attack animation does not work as intended and I would really like the animations to stop when the enemies die

StortSpelProjekt/Engine/src/ECS/Components/AnimationComponent.h
LeoWikstrom: Not a pointer, should not have the p

StortSpelProjekt/Engine/src/Renderer/Animation.h
```diff
@@ -18,10 +18,22 @@ struct TransformKey
...
 	DirectX::XMFLOAT3 position = {};
 	DirectX::XMFLOAT3* pPosition = &position;
-	DirectX::XMFLOAT4 rotationQuaternion = {};
-	DirectX::XMFLOAT4* pRotation = &rotationQuaternion;
+	DirectX::XMFLOAT4 rotation = {};
+	DirectX::XMFLOAT4* pRotation = &rotation;
 	DirectX::XMFLOAT3 scaling = { 1.0f,1.0f,1.0f };
 	DirectX::XMFLOAT3* pScale = &scaling;
+
+	//TransformKey& operator= (const TransformKey& other)
```
LeoWikstrom: Remove comment if not needed

StortSpelProjekt/Engine/src/Renderer/Animation.h
```diff
@@ -51,6 +63,14 @@ struct SkeletonNode
 	DirectX::XMFLOAT4X4 inverseBindPose;	// Bone offset
 	TransformKey* currentStateTransform;
 
+	SkeletonNode()
```
LeoWikstrom, line 54: Why is this a struct and not a class?

StortSpelProjekt/Engine/src/Renderer/DX12Tasks/ForwardRenderTask.h
```diff
@@ -3,16 +3,16 @@
 
 #include "RenderTask.h"
 
-class FowardRenderTask : public RenderTask
+class ForwardRenderTask : public RenderTask
```
LeoWikstrom, line 6: 😞

LeoWikstrom (changes requested):
A few requests

StortSpelProjekt/Game/src/Gamefiles/Components/AiComponent.cpp
```diff
@@ -568,6 +612,12 @@ void component::AiComponent::updateRange(double dt)
 			{
 				if (m_IntervalTimeAccumulator >= m_AttackInterval && m_DistanceToPlayer >= m_AttackingDistance / 8.0f)
 				{
+					if (m_ShouldPlayAttackAnimation)
```
LeoWikstrom: After testing, I found that the actual shot happens noticeably after the enemy moves his arm. I would suggest checking in this if-statement whether m_SpeedTimeAccumulator is larger than something like m_AttackSpeed / 2. You'll have to test and see which number works best, but something like that should give a more desirable result

LeoWikstrom (dismissed):

fredrix96 (approved):
No memory leaks, demons attacks work, spiders animation work as well, although a little slow, no crash after a few rounds of playing. Life goes on, we are done!

OtakuFreak13 (approved):
Animations looks good enough for our game and quality standard.

PR #341, opened 2020-12-10
Pause Menu Change

AntonAsbrink:
Changed the pause buttons to be shorter and changed the text to be a bit more themematic.

Removed the exit button making it so the player has to return to main menu to exit.

Added a warning text when Abandoning(returning to menu) that the player will lose progress.

Made a variable for moving the entire upgrade window in pause menu without losing each components relative position.


LeoWikstrom (dismissed):
The options in the menu are not aligned with each other. The buttons in the Upgrade menu are not updated, and does not show their marked textures. If this can be fixed for the menu buttons, it should be fixable also for these buttons

jakkeez (approved):
Menu works fine and the new theme is good for the game.

LeoWikstrom (approved):

PR #342, opened 2020-12-10
Gameplay Stat Tracking

AntonAsbrink:
This feature tracks what players buy at what level, how long the levels last, how long the game last etc..
This can be toggled in config if you dont want a bunch of files.
I would like if you play serious games to have tracking on and send me the files that gets saved in the game folder so i can make some statistics about the game.


OtakuFreak13 (approved):
Seems like it works as intended. Guessing the time differance between time on levels and total time is because of time spent in the shop scene.

OtakuFreak13:
Just a question not a change request but is it possible to write a file also when you click "Abandon" in pause menu?

LeoWikstrom (approved):

PR #344, opened 2020-12-10
Flickering Lights

fjz345:
Lights have flickeringRate & flickeringAmplitude which controlls a flicker() function.

FLAG::FLICKER is used to flicker

In game:
Torches & braisers have flicker.
Chandileer have lower flicker rate and lower flicker amplitude
Player shots have flicker
enemy shots have flicker


Whixx (approved):

fredrix96 (approved):
The scenes look more lively now. Good job!

PR #345, opened 2020-12-10
memLeakFix

Schtekt:
Turns out an early return caused all this headache.


jakkeez (approved):

fredrix96 (approved):
Seems fine, good job!

PR #348, opened 2020-12-11
Temp All GUI hiding

fredrix96:
GUIOFF hides all gui
GUION shows all gui


jocke1995 (changes requested):
Put everything in a #define, so that we can easily strip the if-statement from dist-build. just wrap everything inside it. Look at the other DEVELOPERMODE defines for tips 👍

LeoWikstrom (changes requested):
The hiding works, but I have two objections: I don't like adding a command to ImGUI without adding it to the recognised commands (so it shows up in HELP and autocomplete), though I guess this is just temporary... But, more importantly, I fail to see the reason to use ImGUI for this since it means we won't be able to remove ALL GUI from the screen. A keybind would be better

fredrix96:
Okay, so I tried a new approach to meet both of your requests. You can now choose if the user will be able to hide or show the GUI and ImGUI by pressing TAB. The TAB button is activated by changing the dev-mode values in core.h

jocke1995 (approved):

LeoWikstrom (approved):
Fint 👍🏽

PR #349, opened 2020-12-13
Particles are more evolved.

fjz345:
Features added:
Color
ColorByLifetime
SizeByLifetime

Tweaked all particle effects.
Tweaked the rooms.


StortSpelProjekt/Engine/src/Particles/ParticleEffect.cpp
```diff
@@ -188,93 +194,59 @@ void ParticleEffect::initParticle(Particle& particle)
...
 }
 
-void ParticleEffect::randomizeRotation(Particle& particle)
+float ParticleEffect::randomizeFloat(float lower, float upper) const
 {
-	RandomParameter randParam = m_Settings.randRotation;
-
-	if (!(randParam.intervalLower == 0 && randParam.intervalUpper == 0))
+	// rand.Randf can't handle same numbers
+	if (lower == upper)
```
LeoWikstrom, line 232: Could be bad to directly compare two floats. Better to use the EPSILON define

LeoWikstrom (changes requested):
The new particle effects look fantastic. However, they do not seem to appear on the chandelier in the Spawn Room and the Empty Room. Also, the OnDeath particles for enemies seem to have disappeared...

LeoWikstrom (approved):

Schtekt (approved):
Goodie, would like to see the particles of the brasier scaled smaller a bit but that is not a big issue.

PR #350, opened 2020-12-14
FreeCam

LeoWikstrom:
Re-added free cam. Press left CTRL to toggle


AntonAsbrink (approved):
Very stuttery please don't rotate

jakkeez (approved):
As Anton stated, its wonky if you move it, but to capture still angles its fine as it is.

PR #351, opened 2020-12-14
Gave Enemies Courage

LeoWikstrom:
Enemies now get less and less shy the longer they exist


jakkeez (approved):

AntonAsbrink (approved):
Appears to work, looks very funny when zombies suddenly see player

PR #352, opened 2020-12-14
Enemy Scaling

AntonAsbrink:
Change how enemies scaled a bit to give a slight increase over time.
There is now a header called EnemyStatDefined that holds all enemies base values and how they scale, giving easy access for anyone want to test or future changes.

There is a new variable called DifficultyScaler that dictates the hp and damage of enemies. This values is currently increased by 0.5 every second there is an enemy and 200 every level. The increase over time corresponds to about an extra level every 4-6 level depending on the speed of the player. This should give a small increase of difficulty of the game that gets more significant the longer the game goes on for, since the level time increases.

The enemies stats was adjusted a bit, making early game easier. specifically the zombies hp starts at 50 meaning the player one shots on the first level


StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.h
```diff
@@ -138,7 +153,7 @@ class EnemyFactory
 	std::map<std::string, EnemyComps*> m_EnemyComps;
 	std::vector<Entity*> m_Enemies;
 	std::vector<float3> m_SpawnPoints;
-	EngineRand m_RandGen;
+	EngineRand* m_RandGen;
```
Schtekt: Naming convention, add p for pointer.
m_RandGen => m_pRandGen

StortSpelProjekt/Game/src/Gamefiles/EnemyFactory.cpp
```diff
@@ -362,6 +365,21 @@ EnemyComps* EnemyFactory::DefineEnemy(const std::string& entityName, EnemyComps*
 	return enemy;
 }
 
+EnemyComps* EnemyFactory::GetDefineEnemy(std::string enemyName)
+{
+	if (m_EnemyComps.find(enemyName) == m_EnemyComps.end())
+	{
+		Log::PrintSeverity(Log::Severity::CRITICAL, "Tried to acess a unexisted enemy comp " + enemyName);
```
LeoWikstrom, line 372: Tried to access an enemy comp that does not exist

LeoWikstrom (approved):
It's fine

Schtekt (approved):
good job

PR #354, opened 2020-12-14
Removed textures and fixed imgui for ram-usage

jocke1995:
Look at the ram usage in ImGui and/or  in the task manager. You should see that as soon as a texture is seen in the scene (uploaded to VRAM), it is removed from RAM.


AntonAsbrink (approved):

LeoWikstrom (approved):
It works

PR #357, opened 2020-12-15
Brightness options, smaller health gui, crosshair in shop

fredrix96:
You can change brightness between 0 and 5 in options


OtakuFreak13 (approved):
The brightness option works great! Looks really good both when lowering or making it brighter.

Schtekt (approved):
Nothing to complain on. Great work!

PR #360, opened 2020-12-16
No thread limit and a true config file

fredrix96:
Unlimited thread power! Oh, also, the config file is now a cfg file instead of a txt file


StortSpelProjekt/Vendor/config.cfg
```diff
@@ -16,4 +16,4 @@ f_sensitivityX 1.0
 f_sensitivityY 1.0
 i_music 1
 i_trackGame 1
-f_brightness 1.0
+f_brightness 2.0
```
kattalainen: I think you forgot to change this back :>

kattalainen (changes requested):

kattalainen (approved):
After playing a bit brightness 2.0 actually feels more nice :> sry I made you change it. Feel free to change it back if you want I'm approving anyway.

Schtekt (approved):
I see no problems, my computer actually got a good performance boost by this :D
Good initiative!

PR #362, opened 2020-12-16
Options menu volume change

kattalainen:
Made the volume option to change master volume instead of only the 3d sounds. Now when volume is 0 you will hear nothing (before you would still hear 2d sounds). The volume changes also don't require game restart anymore. 
The music option will now also stop the menu music when unchecked.


Schtekt (dismissed):
Looks good in general, however I am confused as to why you kept the regular "volume" variable in the config file. Since you only change the master volume in options, would that variable not be of use any more?
Even so, there are places in the code that gets this variable. (AudioVoice.cpp line 35, Audiobuffer.cpp line 183).

kattalainen:
Tim I kept the old volume variables in case someone (including us) wanted to adjust those levels anyway. I could change them to "internal" variables instead but I don't really see a problem having them in the cfg file?

jocke1995 (dismissed):
music doesn't stop when pressing the music-box. Try to fix, if its not possible. then skip it.

jocke1995 (approved):

Schtekt (approved):

PR #363, opened 2020-12-16
Xbox Controller support!

Whixx:


fredrix96 (changes requested):
Would be good to add an option (bool) to be able to choose between a controller or mouse and keyboard. Because now the game is unplayable if the controller is connected or in my case (wireless) powered on.

fredrix96 (approved):
Seems to work now! It feels so great to play with a controller

kattalainen (approved):

PR #364, opened 2020-12-17
ProgressBar stuttering fix

jocke1995:
Look at the progressbars, if they dont "lagg", the fix works. (Not the problem were the progressbar sometimes blinks red)


Schtekt (approved):
Goodie!

LeoWikstrom (approved):

PR #366, opened 2020-12-17
OBB and collision volume adjustments.

Schtekt:
The bullet will no longer bounce of off objects when it does not have bounce upgrade.


jakkeez (approved):

LeoWikstrom (approved):
I'm sure it works fine 😘

PR #370, opened 2020-12-21
Credit scene and a better GUI system

fredrix96:
Play some rounds and see if the works as it should. If you want to try debug you have to set DEVELOPERMODE_DEVINTERFACE i Core.h to true


LeoWikstrom (changes requested):
Controller input does not work in credits scene. To fix this, go to row 203 in Input.cpp and add the creditsScene to the other scenes. Also, tried to check for memory leaks, but the program crashes in debug mode...

LeoWikstrom (approved):

OtakuFreak13 (approved):
Runs as intended and code seems fine.

